PR #1, opened 2017-06-06
[ADD] account_asset_maintenance

pedrobaeza:
Link between assets and equipments
==================================

This module links assets with equipments, allowing to create automatically
an equipment from the supplier invoice if an equipment category is assigned
in the asset category.

Configuration
=============

To configure this module, you need to:

* Go to *Accounting > Configuration > Management > Asset Types*.
* Select one asset type.
* Fill the field "Equipment category" with the equipment category we want to
  assign to the auto-created equipments.

Usage
=====

To use this module, you need to:

* Go to *Accounting > Purchases > Vendor Bills*.
* Create a new bill.
* Create one invoice line.
* Select an asset category with an equipment category filled.
* Validate the invoice.
* A new page called "Equipments" will appear with the auto-created equipments.
* An equipment will created per each quantity indicated in the invoice line.
* If the quantity is not integer (for example: 3.5), the upper integer number
  will be used (4 for the example).
* If you cancel the bill, the created equipments will be removed.

You can access equipments for the created asset:

* Go to *Accounting > Adviser > Assets*.
* Open the created asset
* You will see an smart-button with the string "Equipments" that links to the
  created asset.

### DIFF
--- a/account_asset_maintenance/__manifest__.py
+++ b/account_asset_maintenance/__manifest__.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Link between assets and equipments",
+    "summary": "Create equipment when validating an invoice with assets",
+    "version": "10.0.1.0.0",
+    "category": "Accounting & Finance",
+    "website": "https://www.tecnativa.com/",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "account_asset",
+        "account_cancel",
+        "maintenance",
+    ],
+    "data": [
+        "views/account_asset_asset_views.xml",
+        "views/account_asset_category_views.xml",
+        "views/account_invoice_views.xml",
+        "views/maintenance_equipment_views.xml",
+    ],
+}
--- a/account_asset_maintenance/models/account_asset_asset.py
+++ b/account_asset_maintenance/models/account_asset_asset.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountAssetAsset(models.Model):
+    _inherit = 'account.asset.asset'
+
+    equipment_ids = fields.Many2many(
+        comodel_name="maintenance.equipment", string="Equipments",
+    )
+    equipment_count = fields.Integer(
+        string="Equipment count", compute="_compute_equipment_count",
+    )
+
+    @api.depends('equipment_ids')
+    def _compute_equipment_count(self):
+        for asset in self:
+            asset.equipment_count = len(asset.equipment_ids)
+
+    @api.multi
+    def button_open_equipment(self):
+        self.ensure_one()
+        res = self.env.ref('maintenance.hr_equipment_action').read()[0]
+        res['domain'] = [('asset_ids', 'in', self.ids)]
+        res['context'] = {'default_asset_ids': [(6, 0, self.ids)]}
+        return res
--- a/account_asset_maintenance/models/account_invoice.py
+++ b/account_asset_maintenance/models/account_invoice.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+import math
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    equipment_ids = fields.Many2many(
+        comodel_name="maintenance.equipment", compute="_compute_equipment_ids",
+        string="Equipments",
+    )
+
+    @api.multi
+    @api.depends('invoice_line_ids', 'invoice_line_ids.equipment_ids')
+    def _compute_equipment_ids(self):
+        for invoice in self:
+            invoice.equipment_ids = [
+                (6, 0, invoice.mapped('invoice_line_ids.equipment_ids').ids),
+            ]
+
+    @api.multi
+    def action_invoice_cancel(self):
+        res = super(AccountInvoice, self).action_invoice_cancel()
+        self.mapped('equipment_ids').unlink()
+        return res
+
+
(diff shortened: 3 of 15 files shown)


pedrobaeza:
cc @Tecnativa

account_asset_maintenance/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Link between assets and equipments",
+    "summary": "Create equipment when validating an invoice with assets",
+    "version": "10.0.1.0.0",
+    "category": "Accounting & Finance",
```
dreispt, line 9: Maintenance?

rafaelbn (approved):
Tested in runbot 👍

account_asset_maintenance/__manifest__.py
rafaelbn: @dreispt this module help accountants, I think is Accounting & Finance IMHO

account_asset_maintenance/models/account_invoice.py
```diff
@@ -0,0 +1,68 @@
...
+            })
+        return vals_list
+
+    @api.multi
+    def asset_create(self):
+        for line in self.filtered('asset_category_id.equipment_category_id'):
+            # Create equipments
+            equipments = self.env['maintenance.equipment']
+            for vals in self._prepare_equipment_vals_list(line):
+                equipments += equipments.create(vals)
```
dreispt, line 61: Why don't we create the Equipment on the `account.asset.asset` create method?
That would solve the hack below, that is resource inefficient.

pedrobaeza: Because you don't want an equipment on all created assets. You only want it when created through the invoice with the corresponding data. And also with your proposal we will need a hack to know which is the source invoice line.

dreispt: I want to have an Equipment for every Asset belonging to certain [Asset Categories](https://github.com/OCA/maintenance/pull/1/files#diff-eb09133d34676326e4e5c593a269d519R57).
And new Assets don't necessarily come from supplier invoices.
They may result from an the merge of another company, from the split/merging of existing assets.
Or, in my case, Assets come out of internal installation Projects. The parts bought have invoices, but the final "Installed system" doesn't. Is is then rented/leased to the customer (rather than sold), so it is a company Asset rather than a Sold Product.

pedrobaeza: Well, that's not the use case of this module, sorry.

dreispt: But no point in blocking it either.
Why not creating the Equipments directly in the Asset creation logic.
IMO it makes more sense to be like that.

pedrobaeza: Please write then a PR to my branch changing that, but please respect the link to the account invoice line.

dreispt: Assets should keep a link the origin Invoice Line, but I don't see the point to have that on the Equipment.

pedrobaeza: There is logic for removing the equipment when canceling the invoice line for example, and I want that link anyway. Again, the logic is implemented for starting the flow from the invoices, not from the assets. You want a total different thing that is not the purpose of this module.

dreispt: That logic is between Invoices and Assets; Equipments should be "just" a side effect.
If you cancel an Invoice Line, it should cancel the corresponding Asset.
It should then be the Asset object's job to make sure that the cancellation also happens on the linked Equipments. 
Isolation of concerns.
AFAICS Equipments have no need to be linked to Invoices, only to Assets.

pedrobaeza: OK, please make the PR

dreispt: I don't plan to work on a PR.
I just think the implementation should leave the way open for other use cases.
I think it's reasonable to leave the logic in the Invoice Lines for now, as long as its possible to move it later to the Asset in a stable way.
For that, the Equipment should be independent from Invoice Lines.
That means removing the `invoice_line_id`. Is that possible?

pedrobaeza: Well, if you are not going to work on this, there's no more budget to work ourselves on changing it. Anyway, the field should continue and be filled when possible.

account_asset_maintenance/models/account_asset_asset.py
```diff
@@ -0,0 +1,29 @@
...
+        string="Equipment count", compute="_compute_equipment_count",
+    )
+
+    @api.depends('equipment_ids')
+    def _compute_equipment_count(self):
+        for asset in self:
+            asset.equipment_count = len(asset.equipment_ids)
+
+    @api.multi
+    def button_open_equipment(self):
```
espo-tony, line 24: Wouldn't it be nicer to directly open the form view if there is only one record to show?

account_asset_maintenance/models/account_invoice.py
dreispt: OK, that's perfectly reasonable. My worries is that we should make sure that, if `invoice_line_id` is not there (empty or field removed) it won't break anything.

pedrobaeza: I don't think so. Field is not required.

dreispt (approved):
LGTM.
Just note that the Equipment `invoice_line_id` field should be considered optional and for information only.

account_asset_maintenance/models/account_invoice.py
rigo1985: Hi Pedro I consider that part (if not all) the points that are raised is resolved with the inheritance by the delegation of the Asset in the Equipment.

pedrobaeza: Well, not really, as you still need to develop the code for creating the asset.

rigo1985: @pedrobaeza in my humble opinion you have to make several modifications to the code, but then it would fulfill what you propose and also what comments @dreispt. Among the modifications should be add a check mark in the category of the Asset which are Equipment. So we could buy an Apartment and enter as an Asset and not as a Equipment. You should create a relationship between the Category of the Equipment and the Category of the Asset for when you want to register a Equipment (which because of the inheritance is already created as an Asset) that I have not purchased. Other changes would have to be made, but I think the module would be a bit more complete.

pedrobaeza: As said, we don't have more budget for this, so if you want to work on this, go ahead.

dreispt:
Nothing blocking this -> merging.

PR #2, opened 2017-06-09
[ADD] maintenance_plan module

grindtildeath:
================
Maintenance Plan
================

This module extends the functionality of Odoo Maintenance module by allowing
an equipment to have different preventive maintenance kinds.

Installation
============

Install the module.

Usage
=====

Instead of defining a period and duration for only one preventive maintenance
per equipment, you can define multiple preventive maintenance kind for each
equipment.

Maintenance Kinds have to be defined through the configuration menu. Their name
have to be unique and can be set as active or inactive, should these not be
used anymore.

On any equipment over the maintenance tab, the maintenance plan will appear
as an embedded list view, allowing to add different maintenance kind with their
own period and duration. The next maintenance date will then be computed
automatically according to today's date and the period defined, but the
maintenance request won't be created automatically as is the case in Odoo's
Maintenance module.

Instead, this module uses the original Cron job of Odoo's Maintenance module
to generate maintenance requests, should there not be any requests which is not
done, at the request date matching the maintenance plan next_maintenance_date
for this equipment and this maintenance kind !

### DIFF
--- a/maintenance_plan/__manifest__.py
+++ b/maintenance_plan/__manifest__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{'name': 'Maintenance Plan',
+ 'version': '10.0.1.0.0',
+ 'author': 'Odoo Community Association (OCA), Camptocamp SA',
+ 'license': 'AGPL-3',
+ 'category': 'Maintenance',
+ 'website': 'http://www.camptocamp.com',
+ 'images': [],
+ 'depends': [
+     'maintenance',
+     ],
+ 'data': [
+     'security/ir.model.access.csv',
+     'views/maintenance.xml'
+     ],
+ 'demo': [
+     'data/demo_maintenance_plan.xml'
+ ],
+ 'installable': True,
+ 'auto_install': False,
+ }
--- a/maintenance_plan/__manifest__.py
+++ b/maintenance_plan/__manifest__.py
@@ -2,6 +2,7 @@
 # Copyright 2017 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {'name': 'Maintenance Plan',
+ 'summary': 'Extends preventive maintenance planning',
  'version': '10.0.1.0.0',
  'author': 'Odoo Community Association (OCA), Camptocamp SA',
  'license': 'AGPL-3',
--- a/maintenance_plan/__manifest__.py
+++ b/maintenance_plan/__manifest__.py
@@ -18,6 +18,7 @@
  'demo': [
      'data/demo_maintenance_plan.xml'
  ],
+ 'post_init_hook': 'post_init_hook',
  'installable': True,
  'auto_install': False,
  }
--- a/maintenance_plan/hooks.py
+++ b/maintenance_plan/hooks.py
@@ -30,10 +30,11 @@ def post_init_hook(cr, registry):
                 r.request_date == equipment.next_action_date)
             if len(request) > 1:
                 raise UserError(_(
-                    "You have multiple preventive maintenance requests on an "
-                    "equipment's next action date. Please leave only one "
-                    "request as preventive on the date of equipment's next "
-                    "action to install the module."))
+                    "You have multiple preventive maintenance requests on "
+                    "equipment %s next action date (%s). Please leave only "
+                    "one preventive request on the date of equipment's next "
+                    "action to install the module."
+                ) % (equipment.name, equipment.next_action_date))
             elif len(request) == 1:
                 request.write({
                     'maintenance_kind_id': maintenance_kind.id,
--- a/maintenance_plan/hooks.py
+++ b/maintenance_plan/hooks.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+from odoo import SUPERUSER_ID, api, _
+from odoo.exceptions import UserError
+
+
+def post_init_hook(cr, registry):
+
+    logging.getLogger('odoo.addons.maintenance_plan').info(
+        'Migrating existing preventive maintenance')
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+
+    equipments = env['maintenance.equipment'].search([('period', '!=', False)])
+
+    if equipments:
+
+        maintenance_kind = env['maintenance.kind'].create({
+            'name': 'Install',
+            'active': True,
+        })
+
+        for equipment in equipments:
+            request = equipment.maintenance_ids.filtered(
+                lambda r: r.maintenance_type == 'preventive' and not
+                r.stage_id.done and
+                r.request_date == equipment.next_action_date)
+            if len(request) > 1:
+                raise UserError(_(
+                    "You have multiple preventive maintenance requests on an "
+                    "equipment's next action date. Please leave only one "
+                    "request as preventive on the date of equipment's next "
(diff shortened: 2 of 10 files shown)


dreispt:
LGTM; please fix TravisCI

Suggestions/questions
- It would be nice to have an action/button on "Equipment" to "Generate Next Maintenance"; the corresponding method would reused by the cron job.
- Wouldn't month duration be better for the maintenance intervals? Either way, it could be a nice addition to also have a duration UoM selector - "Days" or "Months".

grindtildeath:
@dreispt 
Thanks for your suggestions.

- We didn't code the 'Generate next maintenance' button as it would be used only at the creation of a new plan and the idea was to create all the maintenance plans and then run the ir.cron manually, should the requests be generated before it runs automatically. 
- A duration UoM could also be useful, but we wanted to keep it as simple as possible to begin with. 

IMO both could be nice addition to the module, but I don't have time to spend on these features at the moment.

dreispt:
@grindtildeath That's why I mentioned them as suggestion ideas.

dreispt:
Does this have any conflict with the standard module's planned maintenance generation, if both are configured?
Does it disable it or will both will work?

Or from a user PoV:
 have planned maintenance configured on my Equipments, and I then install this. 
What is the impact and what do I need to take care to get everything right?

grindtildeath:
@dreispt 
I added a post-init hook, with the description in README.rst of what will happen and what should be done before installing should the standard feature already bu used.
Feel free to let me know what could be improved.

maintenance_plan/models/maintenance.py
```diff
@@ -0,0 +1,219 @@
...
+            'maintenance_type': 'preventive',
+            'owner_user_id': self.owner_user_id.id or self.env.user.id,
+            'technician_user_id': self.technician_user_id.id,
+            'maintenance_team_id': self.maintenance_team_id.id,
+            'maintenance_kind_id': maintenance_plan.maintenance_kind_id.id,
+            'duration': maintenance_plan.duration,
+        })
+
+    @api.model
+    def _cron_generate_requests(self):
```
dreispt: If this is to be used by a Cron it is an external API, so the name method should not start with an underscore.
Furthermore, I can see this used in a Button somewhere, so I suggest renaming it to "do_generate_requests" and expand the docstring to explain that ii can bu used by a cron job.

maintenance_plan/views/maintenance.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- maintenance.kind : views -->
+    <record id="maintenance_kind_view_search" model="ir.ui.view">
+        <field name="name">maintenance.kind.search</field>
+        <field name="model">maintenance.kind</field>
+        <field name="arch" type="xml">
```
dreispt: This view is redundant - Odoo does name searches by default.

maintenance_plan/models/maintenance.py
```diff
@@ -0,0 +1,219 @@
...
+    @api.depends('maintenance_plan_ids')
+    def _compute_team_required(self):
+        for equipment in self:
+            equipment.maintenance_team_required = len(
+                equipment.maintenance_plan_ids) >= 1
+
+    def _create_new_request(self, maintenance_plan):
+        self.ensure_one()
+        self.env['maintenance.request'].create({
+            'name': _('Preventive Maintenance (%s) - %s') %
```
dreispt: Please extract the dict creation into it's own `_prepare_new_maintenance_request` method, so that it is easy to extend it.

dreispt (commented):
LGTM but made a few comements

maintenance_plan/models/maintenance.py
grindtildeath: Thanks for your review. 
However, as this method is called by standard ir.cron from maintenance module, I suppose it's better to leave its name as it is, so we don't have to update the existing ir.cron : 
https://github.com/odoo/odoo/blob/10.0/addons/maintenance/data/maintenance_cron.xml#L10

dreispt: I see, OK then.

dreispt (approved):

maintenance_plan/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{'name': 'Maintenance Plan',
+ 'version': '10.0.1.0.0',
```
elicoidal: a summary would be welcome

maintenance_plan/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+            request = equipment.maintenance_ids.filtered(
+                lambda r: r.maintenance_type == 'preventive' and not
+                r.stage_id.done and
+                r.request_date == equipment.next_action_date)
+            if len(request) > 1:
+                raise UserError(_(
+                    "You have multiple preventive maintenance requests on an "
+                    "equipment's next action date. Please leave only one "
+                    "request as preventive on the date of equipment's next "
+                    "action to install the module."))
```
elicoidal: Nice to have: indicate which record are faulty

maintenance_plan/models/maintenance.py
```diff
@@ -0,0 +1,222 @@
...
+
+    _name = 'maintenance.kind'
+    _description = 'Maintenance Kind'
+
+    name = fields.Char('Name', required=True, translate=True)
+    active = fields.Boolean('Active Kind', required=True, default=True)
+
+    _sql_constraints = [
+        ('name_uniq', 'unique (name)',
+         "Maintenance kind name already exists !")]
```
elicoidal: remove exclamation mark

maintenance_plan/models/maintenance.py
```diff
@@ -0,0 +1,222 @@
...
+                ('maintenance_type', '=', 'preventive'),
+                ('maintenance_kind_id', '=', plan.maintenance_kind_id.id),
+                ('stage_id.done', '=', True),
+                ('close_date', '!=', False)], order="close_date desc", limit=1)
+            if next_maintenance_todo and last_maintenance_done:
+                next_date = next_maintenance_todo.request_date
+                date_gap = fields.Date.from_string(
+                    next_maintenance_todo.request_date) - \
+                    fields.Date.from_string(last_maintenance_done.close_date)
+                # If the gap between the last_maintenance_done and the
```
elicoidal, line 73: This look like specific rule that might not apply everywhere

maintenance_plan/models/maintenance.py
```diff
@@ -0,0 +1,222 @@
...
+                                  'yet. You should either set the request as '
+                                  'done, remove its maintenance kind or '
+                                  'delete it first.') % (
+                    plan.maintenance_kind_id.name, plan.equipment_id.name))
+        super(MaintenancePlan, self).unlink()
+
+    _sql_constraints = [
+        ('equipment_kind_uniq', 'unique (equipment_id, maintenance_kind_id)',
+         "You cannot define multiple times the same maintenance kind on an "
+         "equipment maintenance plan!")]
```
elicoidal: please remove exclamation mark.

elicoidal (changes requested):
some details and one question on contraint

maintenance_plan/models/maintenance.py
grindtildeath: @elicoidal Thanks for your review.

This code is an adaptation from odoo to keep the behaviour as close as possible to standard module : https://github.com/odoo/odoo/blob/10.0/addons/maintenance/models/maintenance.py#L165

dreispt:
@elicoidal can you review the changes please?

grindtildeath:
@elicoidal @pedrobaeza Can we merge this ?

pedrobaeza (approved):
@elicoidal's concern is answered, so I approve this and merge. Thanks all.

PR #4, opened 2017-06-09
[10.0][ADD] Module maintenance_equipment_scrap

espo-tony:
This module improves the action of scrapping an equipment, sending a message and automatically setting the scrap date when the action is performed.


maintenance_equipment_scrap/README.rst
```diff
@@ -0,0 +1,69 @@
...
+
+This module improves the action of scrapping an equipment, sending a
+message and automatically setting the scrap date when the action is performed.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to 'Settings' -> 'Technical' -> 'Email' -> 'Templates' and create a new template you wish to use for equipment scrapping notifications
```
pedrobaeza: Don't you provide a default template?

maintenance_equipment_scrap/models/maintenance_equipment.py
```diff
@@ -0,0 +1,51 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class MaintenanceEquipment(models.Model):
+
+    _inherit = 'maintenance.equipment'
+
+    asset_id = fields.Many2one('account.asset.asset', string='Asset')
```
pedrobaeza: Why adding this field?

pedrobaeza (changes requested):

maintenance_equipment_scrap/models/maintenance_equipment.py
espo-tony: You're perfectly right. It was a mistake.

maintenance_equipment_scrap/README.rst
espo-tony: In my latest commit I provide it.

maintenance_equipment_scrap/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Maintenance Equipments Scrap',
+    'summary': 'Enhance the functionality for Scrapping Equipments',
+    'author': 'Onestein',
```
dreispt: Add OCA

maintenance_equipment_scrap/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'website': 'http://www.onestein.eu',
+    'category': 'Human Resources',
```
dreispt: Maintenance?

maintenance_equipment_scrap/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    equipment_scrap_template_id = fields.Many2one(
+        'mail.template',
+        string='Equipment Scrap Email Template'
+    )
```
dreispt: What do you think about having the Template the a Equipment category level instead (or also)?

maintenance_equipment_scrap/models/maintenance_equipment.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+
+    @api.multi
+    def action_perform_scrap(self):
+        self.ensure_one()
+        action = self.env.ref(
+            'maintenance_equipment_scrap.wizard_perform_equipment_scrap_action'
+        )
+        result = action.read()[0]
+        return result
```
dreispt, line 24: This module is just about sending an email when you click on a button? 
Don't take me wrong but it sounds shallow - a quick Automated Action can do just the same.
Shouldn't we somehow also mark the Equipment as inactive?

maintenance_equipment_scrap/__manifest__.py
espo-tony: I picked the same category of the Odoo base module 'maintenance', just for keeping consistency with it, but it's fine for me to change it. What do you think?

maintenance_equipment_scrap/models/res_company.py
espo-tony: I think it could be a good idea. Maybe the best is adding it on the Equipment category in order to be automatically proposed on the equipment when its category changes?

maintenance_equipment_scrap/models/maintenance_equipment.py
espo-tony: This module "actively" only does two things:
- Sends an email when a scrap is performed, and only if on the equipment is selected a scrapping template;
- Writes the scrap_date field when a scrap is performed.

Still, I think that what really makes this module useful is that it provides a structure for the equipment scrapping functionality. To better explain, consider that in base Odoo, the equipment scrapping functionality just doesn't exist: as far as I can see, an equipment is considered scrapped when the user manually set a scrap date; there is no state change, no active/inactive switch, nothing else that says that a given equipment is scrapped. 
Because of this lack of structure, it's quite tricky to define additional logics to be ran when an equipment is scrapped (such logic as, for exemple, to send an email when it occurs).

This module makes the scrap_date field read-only and provides a procedure for a user to scrap an equipment: it has to click the "scrap" button, to fill a form and finally to submit it. This provides, apart from the obvious functionality of sending an email, an entry-point for adding more and more logic (making the equipment inactive is a possible example, as it could be making some changes on the related asset, or whatever else we can think about, and that could be added in this or in other, more specific-purpose, modules).

What do you think about that?

dreispt: OK, thanks for the explanation. Indeed it is a good step forward.
I'm pretty sure we also need a Stage/State, but I agree that it could be left for a later PR.
In case you are willing to add that, I strongly suggest reusing the oca/server-tools base_stage_state module.

maintenance_equipment_scrap/models/res_company.py
dreispt: Yes, that could work.

maintenance_equipment_scrap/__manifest__.py
dreispt: OK

maintenance_equipment_scrap/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'summary': 'Enhance the functionality for Scrapping Equipments',
+    'author': 'Onestein, Odoo Community Association (OCA)',
+    'website': 'http://www.onestein.eu',
+    'category': 'Human Resources',
+    'version': '10.0.1.0.0',
+    'license': 'AGPL-3',
+    'depends': [
+        'maintenance',
+        'mail',
+        'stock',
```
dreispt: I can see that you added the Stock dependency to be able to add a Settings option.
You thought well, but I think that this is a too heavy dependency.
I would prefer to have it removed, even if that also means dropping the Settings option for now.
ideally, we could have a "maintenance_settings" module, providing the missing Settings menu for the Maintenance app, and then we could add this option there.
What do you think?

espo-tony: I confess I was quite doubtful about using the stock settings for this, deciding for it just because there isn't a default maintenance settings functionality to use nor I thought it was the case to add it within this module.
Therefore I'm in favor of dropping the settings option for now, looking for a further module that provides a Maintenance functionality.

espo-tony:
@dreispt 
As agreed, I removed the settings option altogether with the 'stock' dependency; Moreover I added the mail template also on category level, automatically loading it when the category is changed on the equipment.

@pedrobaeza 
Please, can you check the latest updates?

dreispt (approved):

espo-tony:
@pedrobaeza 
Can I ask you to review the changes you requested on this PR? Thank you in advance

maintenance_equipment_scrap/README.rst
pedrobaeza: Then change this README for telling about that.

astirpe (approved):

elicoidal (approved):
LGTM

elicoidal:
@pedrobaeza I think your comments were attended and we could move forward.

pedrobaeza (approved):

PR #12, opened 2019-07-01
[11.0] [MIG] maintenance_plan

dalonsod:
Superseeds #11 , fixes cron job issue with demo data


maintenance_plan/__manifest__.py
```diff
@@ -1,13 +1,12 @@
-# -*- coding: utf-8 -*-
 # Copyright 2017 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {'name': 'Maintenance Plan',
  'summary': 'Extends preventive maintenance planning',
- 'version': '10.0.1.0.0',
- 'author': 'Odoo Community Association (OCA), Camptocamp SA',
+ 'version': '11.0.1.0.0',
+ 'author': 'Odoo Community Association (OCA), Camptocamp SA, Appcademy SRL',
```
astirpe: I don't think `Appcademy SRL` should be mentioned as "Author" here, as the contribution was a basic porting of the module. The Contributors list was already updated and IMO that would be enough.

astirpe (changes requested):
Tested in Runbot, the issue with the cron and demo data is fixed. Thanks!

maintenance_plan/__manifest__.py
pedrobaeza: I agree. Please put it on contributors only

dalonsod: Done!

astirpe:
@pedrobaeza it seems to me that tests are not executed at all in this V11 branch. In there something wrong in the configuration of this repo?

pedrobaeza:
That's because https://github.com/OCA/maintenance/blob/7e6854e8f25654cd4a357e8877905f3c7ed021ce/maintenance_plan/tests/__init__.py is empty and it's not importing test file.

dalonsod:
OK, one test is not passed yet. I'll fix it

astirpe:
Thanks @pedrobaeza 
@dalonsod thanks, you should be mentioned in the contributors list of the README as well

dalonsod:
@astirpe it seems everything's ok. Could you check?

astirpe (approved):
Tried on Runbot: seems working fine. Thanks!

PR #31, opened 2019-10-04
[12.0][ADD] maintenance_timesheet

dalonsod:
This module adds timesheets to a request, with these features:

- The timesheets will be automatically filled with the request project and task.
- Only request user_id, maintenance team users, and followers will be able to attach timesheets to the request.
- When creating and modifying timesheets, the request must have an undone stage.

This addon requires equipment and request project additions.


dalonsod:
Rebase done, so `maintenance_project_plan` is already included and can be also tested with `maintenance_timesheet`

maintenance_timesheet/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module adds timesheets to a request, with these features:
+
+- The timesheets will be automatically filled with the request project and task.
+- Only request user_id, maintenance team users, and followers will be able to attach timesheets to the request.
+- When creating and modifying timesheets, the request must have an undone stage.
```
astirpe: Tested in Runbot: it's not possible to create a timesheet of a request that is in a done stage. But it's possible to **modify** a timesheet of a request that is in a done stage and this should not be allowed.

maintenance_timesheet/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
...
+
+This addon requires equipment and request project additions.
```
astirpe: I don't think it's necessary to specify this.

maintenance_timesheet/security/maintenance_timesheet_security.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+        <record id="hr_timesheet_rule_request" model="ir.rule">
+            <field name="name">Users are allowed to access timesheets related to a followed request</field>
+            <field name="model_id" ref="model_account_analytic_line"/>
+            <field name="domain_force">[
+                    '|',
+                        ('maintenance_request_id', '=', False),
+                        '&amp;',
```
astirpe: I think this domain can be simplified:



```xml
            <field name="domain_force">[
                    '|',
                        ('maintenance_request_id', '=', False),
                            '|',
                                ('maintenance_request_id.message_partner_ids', 'in', [user.partner_id.id]),
                                ('maintenance_request_id.user_id.id', '=', user.id)
                ]
            </field>
```

maintenance_timesheet/security/maintenance_timesheet_security.xml
```diff
@@ -0,0 +1,23 @@
...
+                            ('maintenance_request_id', '!=', False),
+                            '|',
+                                ('maintenance_request_id.message_partner_ids', 'in', [user.partner_id.id]),
+                                ('maintenance_request_id.user_id.id', '=', user.id)
+                ]
+            </field>
+            <field name="perm_read" eval="1"/>
```
astirpe: The fields perm_read, perm_create, etc.. are by default True, so the can be removed from here.

astirpe (changes requested):
I have some access rights errors in Runbot, here are the steps:

 - create a new user with groups: Timesheet User and Maintenance Full
 - create the employee for that user
 - create a timesheet inside a maintenance request

The strange thing is that I cannot reproduce it locally.
@dalonsod could you also check in Runbot?

dalonsod:
> 
> 
> I have some access rights errors in Runbot, here are the steps:
> 
>     * create a new user with groups: Timesheet User and Maintenance Full
> 
>     * create the employee for that user
> 
>     * create a timesheet inside a maintenance request
> 
> 
> The strange thing is that I cannot reproduce it locally.
> @dalonsod could you also check in Runbot?

Reproduced even locally.

With the current permissions, a "Full Access" user (or, if `base_maintenance_group` module is not installed, "Equipment Manager") can access every request, not only request user and followers.
But I've defined a _global_ access rule which only permits read and write permissions for request user and followers.

What do you think about it, should the rule be a group one, for e.g. "Timesheet user", since global should be too restrictive?

astirpe:
Yes, right! That rule cannot be global. I think "Timesheet user" group is fine for that rule, let's see how it behaves.

Would be also nice to hide the Timesheet button from the request in case the user doesn't have any timesheet group.

dalonsod:
Finally two rules were created, for "Timesheet user" and "Equipment Manager"

maintenance_timesheet/readme/DESCRIPTION.rst
dalonsod: Fixed

dalonsod:
> 
> 
> Would be also nice to hide the Timesheet button from the request in case the user doesn't have any timesheet group.

Applied this suggestion too, for tree, kanban & form views

lmiguens-solvos:
@dalonsod In a Maintenance Request in stage "Repaired" (done) I can delete a timesheet record. 
I think that is not correct, I should not let delete it.

dalonsod:
> 
> 
> @dalonsod In a Maintenance Request in stage "Repaired" (done) I can delete a timesheet record.
> I think that is not correct, I should not let delete it.

Done!!

dalonsod:
@astirpe travis is not passed because of replace made on a view (I don't know why the same code passed previous tests), and for a missing newline I'll fix soon.

Could you take a look for the xpath replace issue in `maintenance_request_views.xml`? I really don't know what should be the best approach to do this trick: move kanban state indicator to the left of the sheet and use the right corner for buttons in the maintenance request form. Thanks!!

dalonsod:
> Could you take a look for the xpath replace issue in `maintenance_request_views.xml`? I really don't know what should be the best approach to do this trick: move kanban state indicator to the left of the sheet and use the right corner for buttons in the maintenance request form. Thanks!!

@astirpe could you help me with this? It's only a dessign issue, but I'm not sure what should be the most convenient approach to fix it. Thanks!

astirpe:
Travis fails for this error:

`maintenance_timesheet/tests/test_maintenance_timesheet.py:73:5: E303 too many blank lines (2)
`
Just remove one blank line where indicated.

Instead, the following:

```
maintenance_timesheet/data/demo_maintenance_timesheet.xml:1: [W7908(missing-newline-extrafiles), ]  Missing newline

maintenance_timesheet/views/maintenance_request_views.xml:55: [W7940(dangerous-view-replace-wo-priority), ]  Dangerous use of "replace" from view with priority 0 < 99

```

are just warning, but it's good to have them fixed. So:

1. add an extra line in demo_maintenance_timesheet.xml
2. try to avoid the use of `position="replace"`. If you cannot avoid it, then just set a high priority (>99) to the inheriting view.

dalonsod:
Thanks, all travis errors and warnings fixed.

astirpe (approved):

lmiguens-solvos (approved):
LGTM 👍

PR #32, opened 2019-10-08
[FIX] maintenance_project: tabs instead of spaces in equipment view form

dalonsod:
This travis flake test was previously passed; anyway I fix it, because other PRs that use `maintenance_project` will fail without this correction


pedrobaeza (approved):
Curious indeed...

PR #35, opened 2019-10-28
[12.0] [FIX] maintenance_plan: next_maintenance_date should be stored

dalonsod:
The current definition of `next_maintenance_date` is incomplete: it has a computed method with `api.depends` decorator, but is not stored. This PR fixes it. 
This change will increase the speed of processes such as cron maintenance request generation.


astirpe (approved):

lmiguens-solvos (approved):
LGTM 👍

PR #36, opened 2019-10-29
[12.0] [IMP] maintenance_plan: make next_maintenance_date field editable

dalonsod:
This PR allows to edit `next_maintenance_date` computed and stored field of the `maintenance_plan` module

### DIFF
--- a/maintenance_plan/models/maintenance.py
+++ b/maintenance_plan/models/maintenance.py
@@ -38,7 +38,8 @@ class MaintenancePlan(models.Model):
 
     next_maintenance_date = fields.Date('Next maintenance date',
                                         compute='_compute_next_maintenance',
-                                        store=True)
+                                        store=True,
+                                        readonly=False)
 
     @api.depends('period', 'maintenance_kind_id',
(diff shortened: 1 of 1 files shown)


dalonsod:
@astirpe I made the following tests:

1. Edit the next maintenance date for an equipment with an existing pending request (eg. Samsung Monitor 15"/MT/122/11112222, on 11/28/2019). Changing the monthly plan date doesn't take effect (I tried 11/25 and 12/05), I suppose the date is automatically re-calculated.
2. The same without any pending request (I deleted the existing one), with the same results.

Could you check? Thanks!!

astirpe:
@dalonsod Yes, I see now. Reproduced in Runbot. Strange indeed, I need time to debug.

astirpe:
I spent some time to debug. There could be a glitch of the ORM, since reading the stored field by using mapped() should not trigger the compute method.
Anyway I think it's not worth to spend further time on this. My original request was meant to improve the UX with just a minimum effort.
@dalonsod  Feel free to close this PR if you don't want to continue.

PR #38, opened 2019-11-11
[12.0][FIX] maintenance_timesheet: python literal text not translatable

dalonsod:
A text defined as variable was wrongly defined in order to be translatable. With this PR the issue is solved

### DIFF
--- a/maintenance_timesheet/tests/test_maintenance_timesheet.py
+++ b/maintenance_timesheet/tests/test_maintenance_timesheet.py
@@ -4,7 +4,6 @@
 import odoo.tests.common as test_common
 from odoo import fields
 from odoo.exceptions import ValidationError
-from ..models import hr_timesheet
 
 
 class TestMaintenanceTimesheet(test_common.TransactionCase):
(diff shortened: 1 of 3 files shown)


astirpe (approved):

lmiguens-solvos (approved):
LGTM 👍

jarroyomorales (approved):

pedrobaeza:
I'm not sure this will work, as the translation is performed at the Odoo start, and thus the language taken will be the initial one, not the desired to be used according user language. The previous form has the problem of not being recognized by the POT extract tool, but works if you inject manually the translation. So, the solution is to not use the string literal at all and put the sentence directly on the code.

pedrobaeza:
Aborting the merge!

astirpe:
OK thanks

astirpe:
@dalonsod I just tried on Runbot and it seems that the translation doesn't work, as Pedro said in https://github.com/OCA/maintenance/pull/38#issuecomment-552785927.

Can you put the sentence directly in the code?

astirpe:
@pedrobaeza is it ok to merge it?

maintenance_timesheet/tests/test_maintenance_timesheet.py
```diff
@@ -55,18 +57,18 @@ def test_check_request_done(self):
                 'user_id': self.env.ref('base.user_admin').id,
                 'date': fields.Date.today(),
                 'unit_amount': 2})]
-        self.assertTrue(hr_timesheet.BAD_TIMESHEET_REQUEST_DONE_MSG
+        self.assertTrue(BAD_TIMESHEET_REQUEST_DONE_MSG
```
pedrobaeza: This is going to fail if the test user language is not English. I recommend you to not check this. It's enough to check for the rise of the exception.

pedrobaeza (approved):

PR #24, opened 2019-08-05
[11.0][add] maintenance_equipment_hierarchy

JordiBForgeFlow:
This module allows to manage a hierarchy of equipments.

image

image

PR #39, opened 2019-11-11
[11.0] [ADD] maintenance_equipment_hierarchy

dalonsod:
This module allows to manage a hierarchy of equipments.

This PR continues #24 , with these improvements:

- Fixes issue when creating an equipment child from another one.
- Adds name_get() method in order to display equipment hierarchy.
- Parent equipment field is moved to the top of the form view.

When this PR is merged, I'll migrate the addon to v12.


etobella (commented):
Could you add the check of the parents on the name_search
Something similar to https://github.com/odoo/odoo/blob/11.0/addons/product/models/product.py#L440

On the other hand. It might be interesting to use the following widget in order to select the parent:
OCA/web#1430

maintenance_equipment_hierarchy/models/maintenance_equipment.py
```diff
@@ -0,0 +1,61 @@
...
+    parent_id = fields.Many2one('maintenance.equipment', 'Parent Equipment',
+                                index=True, ondelete='cascade')
+    child_ids = fields.One2many('maintenance.equipment', 'parent_id',
+                                'Child Equipments')
+    parent_left = fields.Integer('Left Parent', index=1)
+    parent_right = fields.Integer('Right Parent', index=1)
+    child_count = fields.Integer(
+        compute='_compute_child_count',
+        string="Number of child equipments")
+
```
astirpe, line 26:
```suggestion
    display_name = fields.Char(compute='_compute_display_name')

    @api.depends('name', 'parent_id.display_name')  # recursive definition
    def _compute_display_name(self):
        for equipment in self:
            if equipment.parent_id:
                parent_name = equipment.parent_id.display_name
                equipment.display_name = parent_name + ' / ' + equipment.name
            else:
                equipment.display_name = equipment.name

```

This is also needed IMO.

astirpe (changes requested):

maintenance_equipment_hierarchy/models/maintenance_equipment.py
dalonsod: I agree... In fact, if `display_name` field is defined as stored, it's possible to use it in a `_name_search()` implementation, as @etobella suggested.

And `_rec_name='display_name'` should be other improvement, as well

@astirpe what do you think?

astirpe: I don't like the idea to make display_name as stored. I also don't like the idea of  `_rec_name='display_name'`.

I would rather prefer a new dedicated field (eg.: `complete_name` or `full_path` or similar), computed and stored, to be used for the `_name_search()`; also the actual `_compute_display_name()` could benefit of it and be simplified.

Consider also to add field `parent_id` in the equipment listview.

dalonsod: Thanks, I'll make the following changes as soon as possible:
- new `display_name` computed field
- `_name_get()` implementation
- new `complete_name` computed and stored field, which will be used by `_compute_display_name()` and `_name_get()` methods
- add `parent_id` to equipment list view.

dalonsod:
> On the other hand. It might be interesting to use the following widget in order to select the parent:
> [OCA/web#1430](https://github.com/OCA/web/pull/1430)

I'll take a look, thanks

maintenance_equipment_hierarchy/models/maintenance_equipment.py
dalonsod: I've done the changes described above.
However, I didn't implement `_name_search()` because there was already a filter definition in the search view for equipment name, and I adapted that view code instead.
I've also added some demo data.
@astirpe could you check? Thanks!

maintenance_equipment_hierarchy/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,45 @@
...
+            </field>
+        </field>
+    </record>
+
+    <record id="hr_equipment_view_tree" model="ir.ui.view">
+        <field name="name">equipment.tree</field>
+        <field name="model">maintenance.equipment</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_view_tree"/>
+        <field name="arch" type="xml">
+            <field name="category_id" position="after">
```
astirpe: I think it would be much more handy if the `parent_id` is placed just after the Equipment Name.

astirpe (approved):
Just one comment. For the rest, it looks good to me.

etobella (approved):

maintenance_equipment_hierarchy/views/maintenance_equipment_views.xml
dalonsod: Done!

astirpe:
@jbeficent this PR comes from your original work, would you review/merge it?

LoisRForgeFlow (approved):
tested and LGTM :+1: 

@astirpe  I am aware of the project details @jbeficent did this for, so you can proceed with the merge.

Thanks!

PR #41, opened 2019-11-13
[11.0][ADD] maintenance_request_sequence

jarroyomorales:
Adds a sequence field to maintenance requests.


JordiBForgeFlow:
@jarroyomorales Sometimes a company may want to maintain different sequences, depending on the `maintenance.kind`. I think that it could be a good idea to allow a user to create a sequence in this model, that would then be used when creating a maintenance request that used this kind.

Look how it's done in the `product_sequence` module https://github.com/OCA/product-attribute/blob/11.0/product_sequence/models/product_category.py

jarroyomorales:
@jbeficent I thought about that, it is the same than in `maintenance_equipment_sequence` if I'm not wrong.  At first my idea was to create a sequence per team, but it would cause problems when changing the team of the requests. 
I also have problems understanding the usage of kind. I thought it was like a category field to group similar requests, but demo data's examples are 'Weekly' and 'Monthly', which confuses me, since I wouldnt group requests by it's frecuency, it is more interesting to group them by the type of actions they will require or something like that. Maybe the only problem is that demo data is not the best.
Another thing is that kind is not a required field in requests, which sequence would they get? Should we make the field required? Also, maybe a user creates a request and then a technician assigns the kind, so the sequence can't be decided on creation.

JordiBForgeFlow:
@jarroyomorales I agree that the examples of `maintenance.kind` are very confusing. I never use them as 'weekly', or 'monthly', but I use values like:
* Customer complaint
* Internal problem
* Planned maintenance
....

In the end it is the only way to categorize maintenance requests.

etobella:
> In the end it is the only way to categorize maintenance requests.

Well, we can create something like tags in order to categorize, as it would be more helpful, sometimes, maintenance requests could fulfill more than one kind (outside of this module, obviously)

Using Kinds is a bad choice IMO, the module is not related to `maintenance_plan` and kind is not required and can be changed, What happens if we create a MR in a kind and then we change it? Do we need to change the code?

etobella:
Maybe the best options would be to use team as the selector of the sequence and we can maintain the old code if we change the team, at the end we want to know the change of team.

jarroyomorales:
@etobella Now it works how you said, let's see what others think.

maintenance_request_sequence/models/maintenance_request.py
```diff
@@ -0,0 +1,39 @@
...
+                team_id).sequence_id or self.env.ref(
+                'maintenance_request_sequence.seq_maintenance_request_auto'
+            )
+            vals['code'] = sequence.next_by_id()
+        return super().create(vals)
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        # Make a search with default criteria
+        names1 = super(models.Model, self).name_search(
```
etobella: just super().name_search is enough

maintenance_request_sequence/views/maintenance_request.xml
```diff
@@ -0,0 +1,51 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="hr_equipment_request_view_form">
+        <field name="name">maintenance.request.form (in maintenance_request_sequence)</field>
+        <field name="model">maintenance.request</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_request_view_form"/>
+        <field name="arch" type="xml">
+            <field name="owner_user_id" position="before">
```
etobella: Wouldn't be better to add it on the top of the request?

etobella (changes requested):
I think you missed to add a file :wink:

jarroyomorales:
@etobella Comments attended!

etobella (approved):
LGTM :+1:

jarroyomorales:
@jbeficent @ageficent What do you think about this?

JordiBForgeFlow (approved):
:+1: Technical and functional review.

PR #51, opened 2019-12-04
[11.0][ADD] maintenance_request_stage_transition

etobella:
It allows to define the next stages of a stage and shows them on the header buttons.

Usage
=====

* Access `Maintenance > Configuration > Maintenance Stages`
* Define the possible next stages of every stage
* You can specify the color of the button if needed
* Refresh the page in order to show the new possible stages
* The stage widget will be blocked and the expected button will appear on the header.


etobella:
@jarroyomorales

maintenance_request_flow/models/maintenance_request.py
```diff
@@ -0,0 +1,49 @@
...
+from odoo.osv.orm import setup_modifiers
+from lxml import etree
+
+
+class MaintenanceRequest(models.Model):
+
+    _inherit = 'maintenance.request'
+
+    stage_id = fields.Many2one(
+        'maintenance.stage',
```
jarroyomorales: Is it necessary to write the comodel when inheriting the field? Just asking

jarroyomorales (commented):
Tested locally and works as expected :+1: Maybe we could discuss the name, `maintenance_request_stage_transition` or something like that.
Also thought we could make it a mixin, to be able to use it with other similar models, but I dont know how hard can it be...

maintenance_request_flow/models/maintenance_stage.py
```diff
@@ -0,0 +1,55 @@
...
+
+class MaintenanceStage(models.Model):
+
+    _inherit = 'maintenance.stage'
+
+    next_stage_ids = fields.Many2many(
+        'maintenance.stage',
+        relation='maintenance_stage_next_stage',
+        column1='stage_id',
+        column2='next_stage_id',
```
JordiBForgeFlow: string="Next Stages". Otherwise the label will say "Next Stage" which can be misleading.

etobella:
I don't like transition as we are not defining the transitions properly, maybe we could do that, but I am not sure if it is the best approach. :disappointed: 

@eficent what do you think about it? Which option do you prefer?

- Create the transition model (defining the flow and specific functions if wanted)
- Keep this way that maybe is simpler

JordiBForgeFlow:
@etobella @jarroyomorales I think that we could define the module as `maintenance_request_stage_transition`, but implement only the current basic transition mechanism. In the future the module could be extended so that it's possible to indicate in the maintenance settings "Use conditions in transitions between stages".

We could just put this in the roadmap for now.

JordiBForgeFlow (approved):
:+1: Minor comment

AdriaGForgeFlow (approved):
LGTM :+1:

maintenance_request_flow/models/maintenance_request.py
etobella: Not according to Odoo, but I am not sure...

etobella:
@jbeficent @ageficent @jarroyomorales Comments attended :smile:

PR #55, opened 2019-12-13
[11.0][ADD] maintenance_equipment_contract

jarroyomorales:
This module links maintenance equipment with supplier contracts.
To link an equipment go to the contract's form and add it to the Equipments
field.
@etobella


etobella (approved):
LGTM :+1:
Useful when you want to check the related equipments to a maintenance contract.

dalonsod (changes requested):
@jarroyomorales functionally reviewed, looks great, only a few things if possible:
- If an equipment has at least a contract, it's possible to create a contract from the equipment form. But the form provided does not fill the selected equipment as the default one in this new contract (maybe it's difficult to achieve it, I don't know...).
- I think some additional search data should be useful (e.g. equipments in contract view, or contract supplier in equipment view), what you think about it?
- Could you provide some demo data?

jarroyomorales:
@dalonsod Comments attended. Thank you!

dalonsod (approved):
Locally tested, LGTM

PR #58, opened 2019-12-19
[11.0][ADD] maintenance_remote

etobella:
This module allows to show the remote on a maintenance request where the issue
was created. It is useful on requests related to IT.

@jarroyomorales @JordiBForgeFlow


jarroyomorales (approved):
Simple but useful module. LGTM :+1:

dalonsod (approved):
👍 
Not blocking, maybe `remote_id` field should be only visible for a manager role, or when you are in debug mode.

etobella:
Well, a usual user can only see his own issues, but we had the same thought and maybe we can create a specific group. I will check on a few days

jarroyomorales:
Yes, now that you say it I also agree, even if users can only see their requests I think it is not necessary that they see information about their remote. This info is only useful to people that are going to work on the issue.

etobella:
@jarroyomorales @dalonsod I made the changes you commented. I agree this is a better approach

jarroyomorales:
@etobella Wouldnt it be easier to use the  `maintenance.group_equipment_manager` group instead of creating one just for this?

etobella:
Well I decided to make a new group because people from a team may need the remote and other teams don't. But it is probably easier to use `maintenance.group_equipment_manager`

jarroyomorales:
Nice then. Should this PR be merged? @etobella

dalonsod:
I agree 👍

etobella:
\ocabot merge

PR #59, opened 2019-12-24
[11.0][IMP] maintenance_plan: Return the created requests

etobella:
This just returns the requests in order to be inherited properly

PR #19, opened 2019-07-24
[11.0][IMP] maintenance_plan

AdriaGForgeFlow:
Mian improvments included in this PR:

* Extension of maintenance plan module with the addition and renaming of the following fields:
    - Start date
    - Planning Horizon & Plannig Horizon step
    - Instructions page
    - period --> interval
    - Interval setp
* Creation of maintenance requests taking into account planning horizon

Pending:
* [x] Add migration scripts

cc ~ @Eficent

PR #56, opened 2019-12-13
[11.0][FIX] default equipment_id in plans

jarroyomorales:
When creating a plan from equipment's form you had to select the equipment_id again. With this change the field is automatically set and hidden when accessing from the equipment's form.

@etobella @LoisRForgeFlow @JordiBForgeFlow @AdriaGForgeFlow

PR #60, opened 2019-12-24
[12.0] Improve maintenance plan

etobella:
This PR includes the upgrades from
- [x] #19 
- [x] #59 
- [x] Fixes for the existing modules
I know this affects more than one module, but I don't want to have this failing.

@JordiBForgeFlow @LoisRForgeFlow @dalonsod @jarroyomorales


LoisRForgeFlow (approved):
Thanks for forward porting that :+1:

AdriaGForgeFlow (approved):
Looking good! Thanks for the work :+1:

dalonsod (changes requested):
Hello @etobella ,

thanks for `maintenance_project_plan` adaptations, I've tested them and look 👍 

Turning back to `maintenance_plan`,

- there are some travis warnings, could you take a look at them?
- I've found two issues, inherited from v11 implementation I think, maybe they'd be fixed in other PR, what do you think about it? (1) `maintenance.plan` form view display name seems to always show "False", and (2) when creating a plan from `maintenance.equipment` form, the equipment is not filled with the current one (that should also be set as `readonly`).

etobella:
@dalonsod I fixed the travis issues.
About the issues, the problem is the following:
Name is not required, but it is used on the name, maybe we should need to define it as required. What do you think?
The other issue can be solved by setting a default.
I will make a PR on 11 solving this issues and we can cherry-pick the commit here. What do you think?

etobella:
@dalonsod There was already a PR fixing it and it had been merged #56 I already included the commit :smile:

etobella:
Some changes were necessary in order to make it work properly, something changed when computing the invisible on child elements.

@jarroyomorales

jarroyomorales (approved):
LGTM! :+1: Nice porting! But you should also have in mind to cherry pick #62

dalonsod:
> Name is not required, but it is used on the name, maybe we should need to define it as required. What do you think?

I think a `name_get()` method should be enough for this issue, then preserving name field as not required. Anyway, this should be implemented in an independent PR, to be later cherry-picked to v11.

dalonsod (approved):
👍

PR #62, opened 2019-12-24
[11.0][IMP] maintenance_plan: Add team on plans

etobella:
With this module, we can change the default team of a plan. For example, we could have a team for callibration and another one for technical reviews.
@jarroyomorales @dalonsod @LoisRForgeFlow

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -29,11 +29,13 @@ class MaintenanceEquipment(models.Model):
     def _compute_team_required(self):
         for equipment in self:
             equipment.maintenance_team_required = len(
-                equipment.maintenance_plan_ids) >= 1
+                equipment.maintenance_plan_ids.filtered(
+                    lambda r: not r.maintenance_team_id
+                )) >= 1
 
     def _prepare_request_from_plan(self, maintenance_plan,
                                    next_maintenance_date):
-        team = self.maintenance_team_id
+        team = maintenance_plan.maintenance_team_id or self.maintenance_team_id
         if not team:
             team = self.env['maintenance.request']._get_default_team_id()
         return {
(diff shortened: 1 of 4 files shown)


dalonsod (changes requested):
Functionally reviewed,
- `maintenance_team_id` is not shown in embedded `maintenance_plan_ids` tree view within `maintenance.equipment` form.
- also in `maintenance.equipment` form, `maintenance_team_id` becomes required when there's a plan set, at least. You could consider relax that condition depending on the plans (e.g. if every plan actually has their `maintenance_team_id`, the `maintenance.equipment` `maintenance_team_id` shouldn't be set as required).

etobella:
@dalonsod Comments attended

etobella:
Travis is failing due to 
imagen

But OCB is working, I don't know why odoo is not accessible from github...
**https://github.com/odoo/**

maintenance_plan/models/maintenance_equipment.py
```diff
@@ -29,11 +29,13 @@ def _compute_maintenance_plan_count(self):
     def _compute_team_required(self):
         for equipment in self:
             equipment.maintenance_team_required = len(
-                equipment.maintenance_plan_ids) >= 1
+                equipment.maintenance_plan_ids.filtered(
```
jarroyomorales, line 32: Is it more efficient than casting it with `bool()` ?

dalonsod (approved):
LGTM 👍

maintenance_plan/models/maintenance_equipment.py
etobella: When checking bool it is checking if len > 0, so it should be the same

jarroyomorales (approved):
LGTM :+1:

LoisRForgeFlow (approved):
Code review, LGTM

PR #63, opened 2020-01-02
[12.0][FIX] maintenance_plan: prevent False as display name when name is empty

dalonsod:
Since `name` field for `maintenance.plan` is not required, this PR implements a `name_get()` method to prevent `False` to be displayed in the form navigation widget.

E.g. if an "Equipment A" has a "Monthly" kind plan _and name is empty_, "Unnamed Montly plan (Equipment A)" is finally shown.

cc @etobella

### DIFF
--- a/maintenance_plan/tests/test_maintenance_plan.py
+++ b/maintenance_plan/tests/test_maintenance_plan.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import odoo.tests.common as test_common
-from odoo import fields
+from odoo import fields, _
 from datetime import timedelta
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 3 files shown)


jarroyomorales (approved):
LGTM but just a question. Is it OK that tests rely on demo data?

maintenance_plan/tests/test_maintenance_plan.py
```diff
@@ -30,6 +35,20 @@ def setUp(self):
         today = fields.Date.today()
         self.today_date = fields.Date.from_string(today)
 
+    def test_name_get(self):
+        self.assertEqual(
```
etobella, line 51: Can you make the test unrelated to demo data. It is usually a bad plan to make it related to demo data.

etobella (approved):
I think that @jarroyomorales is right about the test data, but the problem is not with the PR, but the test itself, maybe it is time to remake it, but maybe in another PR.

etobella:
@LoisRForgeFlow

maintenance_plan/tests/test_maintenance_plan.py
dalonsod: Done!

LoisRForgeFlow (approved):
LGTM!

PR #72, opened 2020-01-22
[11.0][IMP] maintenance_plan: no equipment

jarroyomorales:
If no equipment was set for a plan, the cron used to fail because of an expected singleton. This PR solves this, allowing you to define plans without equipment.

@etobella @LoisRForgeFlow @NuriaMForgeFlow

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -38,9 +38,15 @@ class MaintenanceEquipment(models.Model):
         team = maintenance_plan.maintenance_team_id or self.maintenance_team_id
         if not team:
             team = self.env['maintenance.request']._get_default_team_id()
+
+        description = self.name if self else maintenance_plan.name
+        kind = maintenance_plan.maintenance_kind_id.name or _(
+            'Unspecified kind'
+        )
+        name = _('Preventive Maintenance (%s) - %s') % (kind, description)
+
         return {
-            'name': _('Preventive Maintenance (%s) - %s') % (
-                maintenance_plan.maintenance_kind_id.name, self.name),
+            'name': name,
             'request_date': next_maintenance_date,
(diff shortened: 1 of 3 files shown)


maintenance_plan/models/maintenance_equipment.py
```diff
@@ -38,9 +38,13 @@ def _prepare_request_from_plan(self, maintenance_plan,
         team = maintenance_plan.maintenance_team_id or self.maintenance_team_id
         if not team:
             team = self.env['maintenance.request']._get_default_team_id()
+
+        description = self.name if self else maintenance_plan.name
+        name = _('Preventive Maintenance (%s) - %s') % (
+            maintenance_plan.maintenance_kind_id.name, description)
```
etobella: maintenance_kind_id is required?

etobella (approved):
LGTM :+1: Just a comment about maintenance_kind

maintenance_plan/models/maintenance_equipment.py
jarroyomorales: No its not, I just fixed this part so it doesn't display `False` on the title.

etobella (changes requested):
Just a question, with this change, we should need to see all the plans in order to configure them. Can you check it please?

jarroyomorales:
@etobella What do you mean? I think there is already a menu for what you are asking. https://github.com/OCA/maintenance/blob/11.0/maintenance_plan/views/maintenance_plan_views.xml#L131

etobella (approved):
Ups!!! Sorry for the misunderstanding, I was unable to find it. My fault :cry:

NuriaXifre (approved):
Functional review done! LGTM

etobella:
@jarroyomorales Can you create the PRs with the cherry-pick for 12.0 and 13.0 please :wink:

PR #76, opened 2020-02-27
[13.0][MIG] maintenance_equipment_hierarchy

dalonsod:
Known issues:

- ~~built-in Odoo recursion checking runs before `_check_equipment_recursion` equipment method, and this one is never executed~~.


dalonsod:
cc @AdriaGForgeFlow

jarroyomorales (commented):
Tested on runbot and works as expected. Regarding the recursion check I think we can remove that function. At first I thought about overriding the function that checks this to change the message but I think its not worth it.
https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L3961

dalonsod:
> 
> 
> Tested on runbot and works as expected. Regarding the recursion check I think we can remove that function. At first I thought about overriding the function that checks this to change the message but I think its not worth it.
> https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L3961

Removed recursion checks

etobella (approved):
LGTM :+1:

JoanSForgeFlow (approved):

PR #87, opened 2020-05-07
[13.0][MIG] maintenance_equipment_scrap

astirpe:


dalonsod (changes requested):
Code and functional review.
I specially tested changes on `equipment_scrap_template_id` computed and editable field, works as expected 👍 

Only one thing: according to migration rules to v13, is it possible to squash certain administrative commits for v12 implementation?

etobella:
@astirpe It works as expected, but I have a few questions.
* Why do you need to edit the scrap template from the equipment?
* Could you edit the default template in order to not write Serial Number False if it is undefined?

astirpe:
> is it possible to squash certain administrative commits for v12 implementation?

@dalonsod all administrative commits for V12 have different authors


> Why do you need to edit the scrap template from the equipment?

@etobella for flexibility of use. I was made that way since the beginning.

etobella:
Thanks!!! If you merge the migration commits we can proceed with merge

astirpe:
Done!

etobella (approved):

PR #94, opened 2020-07-11
[12.0][ADD] Add Maintenance Request Report

marcelsavegnago:
This module adds a report to the maintenance request

### DIFF
--- a/base_maintenance/README.rst
+++ b/base_maintenance/README.rst
@@ -33,6 +33,7 @@ some basic features that are standard in other Odoo apps:
 * Description field in maintenance team.
 * Creates the unexisting search view for maintenance team.
 * Empty button box on maintenance request to be extended by other modules.
+* Maintenance Request Report
 
 **Table of contents**
 
@@ -61,6 +62,8 @@ Contributors
 ~~~~~~~~~~~~
 
 * Lois Rilo <lois.rilo@eficent.com>
+* Marcel Savegnago <marcel.savegnago@escodoo.com.br>
+
 
 Maintainers
 ~~~~~~~~~~~
--- a/base_maintenance/__manifest__.py
+++ b/base_maintenance/__manifest__.py
@@ -16,7 +16,8 @@
     "data": [
         "views/maintenance_team_views.xml",
         "views/maintenance_request_views.xml",
-        "views/maintenance_equipment_views.xml"
+        "views/maintenance_equipment_views.xml",
+        "views/report_maintenance_request.xml",
     ],
     "installable": True,
 }
--- a/base_maintenance/i18n/base_maintenance.pot
+++ b/base_maintenance/i18n/base_maintenance.pot
@@ -6,12 +6,94 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 12.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2020-07-14 23:41+0000\n"
+"PO-Revision-Date: 2020-07-14 23:41+0000\n"
 "Last-Translator: <>\n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: \n"
 "Plural-Forms: \n"
+
+#. module: base_maintenance
+#: model_terms:ir.ui.view,arch_db:base_maintenance.report_maintenance_request_document
+msgid "<span>Maintenance Request</span>"
+msgstr ""
+
+#. module: base_maintenance
+#: model_terms:ir.ui.view,arch_db:base_maintenance.report_maintenance_request_document
+msgid "<strong>Category:</strong>"
+msgstr ""
+
+#. module: base_maintenance
+#: model_terms:ir.ui.view,arch_db:base_maintenance.report_maintenance_request_document
+msgid "<strong>Close Date:</strong>"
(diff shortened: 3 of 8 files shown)


marcelsavegnago:
cc @eduaparicio

maintenance_request_report/views/report_maintenance_request.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 - TODAY, Marcel Savegnago - Escodoo
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <data>
+        <template id="report_maintenance_request_document">
+            <t t-call="web.external_layout">
```
etobella: I think that maintenance requests are usually internal activities, so, we should use internal_layout. What do you think?

maintenance_request_report/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'author': 'Escodoo,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/maintenance',
+    'depends': [
+        'maintenance',
+        'maintenance_request_sequence',
+    ],
+    'data': [
+        'views/report_maintenance_request.xml',
+    ],
+    'demo': [
```
etobella: If no demo, this is not required

maintenance_request_report/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'name': 'Maintenance Request Report',
+    'summary': """
+        Maintenance Request Report""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'category': 'Maintenance',
+    'author': 'Escodoo,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/maintenance',
+    'depends': [
+        'maintenance',
```
etobella: not needed, as maintenance_request_sequence already depends on maintenance

maintenance_request_report/views/report_maintenance_request.xml
```diff
@@ -0,0 +1,129 @@
...
+                <t t-set="doc" t-value="doc.with_context(lang=doc.user_id.partner_id.lang)" />
+
+                <div class="page">
+                    <div class="oe_structure" />
+
+                    <h2 class="mt16">
+                        <t>
+                            <span>Maintenance Request #</span>
```
etobella: I think # shouldn't be required. If you need the # you can modify the sequence, but if you made MR0001, now it will be shown #MR0001 and it has no sense for me

etobella (changes requested):

dalonsod:
Hello! Very interesting addon, I usually miss reports in Maintenance.

@etobella I have a doubt: is there any OCA guideline for addon hierarchy and this kind of addons? E.g. in this case an alternative approach should've been creating this addon only with the base dependency (`maintenance`), not starting from `maintenance_request_sequence`, and maybe later improve `maintenance_request_sequence` and others including the proper inheritance in order to use this report. What do you think?

jarroyomorales:
@dalonsod Yes, it is hard to tell which is the best solution. You dont want to force people use a sequence if they want the report and you dont want to force the report if you want the sequence, but creating a glue module everytime this happens doesnt look very clean to me.

Maybe the best option is to add this report to the `base_maintenance` module with as many things as possible that serves as template and then everyone can inherit it and customize with its own requirements.

dalonsod:
@jarroyomorales trying to improve `base_maintenance` seems to be a good solution IMHO. @marcelsavegnago do you agree?

marcelsavegnago:
> @jarroyomorales trying to improve `base_maintenance` seems to be a good solution IMHO. @marcelsavegnago do you agree?

I do agree. There was even initially thinking about this possibility.

@etobella and @jarroyomorales do you agree?

maintenance_request_report/views/report_maintenance_request.xml
marcelsavegnago: I agree

maintenance_request_report/views/report_maintenance_request.xml
marcelsavegnago: For now I removed the dependency on manitenance_request_sequence

maintenance_request_report/__manifest__.py
marcelsavegnago: For now I removed the dependency on manitenance_request_sequence

maintenance_request_report/__manifest__.py
marcelsavegnago: Done

jarroyomorales:
@marcelsavegnago Yes!

etobella (approved):
LGTM Just one comment, can you change the commit description for something like :
`[IMP] base_maintenance: Add a Maintenance Request Report`

dalonsod (approved):
👍

jarroyomorales (approved):

etobella:
/ocabot merge minor
@marcelsavegnago Can you cherry-pick the commit for 13.0 please???

etobella:
Ocabot failed,
Trying again /ocabot merge minor

marcelsavegnago:
> /ocabot merge minor
> @marcelsavegnago Can you cherry-pick the commit for 13.0 please???

yeap.. 
https://github.com/OCA/maintenance/pull/97

PR #100, opened 2020-08-03
[12.0] Fix write account analytic line with recordset

sergiocorato:
Write is not call correctly when updating many records (like in mass update).
This fix it.


etobella (approved):

etobella:
It is a minor change and an issue, so /ocabot merge patch

PR #104, opened 2020-09-16
[12.0][ADD] maintenance_stock

dalonsod:
This module extends the functionality of Odoo Maintenance module by allowing link stock consumptions to maintenance requests. More information in `README.rst` file


etobella:
Interesting approach! Can you add some tests, please?

dalonsod:
> 
> 
> Interesting approach! Can you add some tests, please?

Thanks! The module has already some tests, I'll check them when travis is OK and then I'll request you for review, comments are welcome

dalonsod:
@etobella increased test coverage. Could you review?
cc @astirpe @marcelsavegnago

marcelsavegnago (commented):
I believe it is better if you change the position of the buttons

https://github.com/OCA/maintenance/blob/0c9cc3f2f21ac2f819f54a02241417c98322993b/base_maintenance/views/maintenance_request_views.xml#L12

I believe there may be some conflict with that, but the right thing would be to change the maintenance_timesheet later to depend on base_maintenance
https://github.com/OCA/maintenance/blob/0c9cc3f2f21ac2f819f54a02241417c98322993b/maintenance_timesheet/views/maintenance_request_views.xml#L60

image

marcelsavegnago (commented):
This is a new PR. You can link the maintenance request to a parts selection list required for this maintenance.

It is very cool and congratulations for the great work, but I believe that you should consider a new page in the equipment register to establish a list of standard products and necessary for maintenance and preventive maintenance. For example, think of a large machine that needs to change its lubricating oil every 30 days.

We can also consider creating a list of standard products for each maintenance plan. Of course, we can put this in the future and in another PR.

dalonsod:
Thanks for your review @marcelsavegnago ,
- The issue with the position of buttons on request form surely is a `maintenance_timesheet` problem. Locally tested, `maintenance_stock` works properly. I actually solved it on v13 migration, but I forgot porting that update to v12 implementation. As soon as possible I'll make the PR to fix it. Meanwhile I think it's not a problem for this PR approval progress, what do you think?
- I agree with you with the proposed improvement of such as "list of standard product list templates" linked to an equipment. Actually, I'd considered including that in a `ROADMAP` file, but I forgot it. I think a good approach for this could be maintaining product standard list templates, and then every equipment could select some of then, because an equipment could need more than one (for different requests), and other equipments could use the same one. With this approach you could manually load a template from the equipment standards template list, or use no one. And, of course, with another addon plans should be linked to this functionality as well. What do you think, adding this functionality right now should be mandatory to approve this addon, or could be a future improvement?

marcelsavegnago:
> * The issue with the position of buttons on request form surely is a `maintenance_timesheet` problem. Locally tested, `maintenance_stock` works properly. I actually solved it on v13 migration, but I forgot porting that update to v12 implementation. As soon as possible I'll make the PR to fix it. Meanwhile I think it's not a problem for this PR approval progress, what do you think?

I agree.

> * I agree with you with the proposed improvement of such as "list of standard product list templates" linked to an equipment. Actually, I'd considered including that in a `ROADMAP` file, but I forgot it. I think a good approach for this could be maintaining product standard list templates, and then every equipment could select some of then, because an equipment could need more than one (for different requests), and other equipments could use the same one. With this approach you could manually load a template from the equipment standards template list, or use no one. And, of course, with another addon plans should be linked to this functionality as well. What do you think, adding this functionality right now should be mandatory to approve this addon, or could be a future improvement?

@dalonsod  I don't think it needs to be mandatory now. So I see no problem dealing with that in the future. The way it is already helps a lot

marcelsavegnago (approved):

dalonsod:
>     * The issue with the position of buttons on request form surely is a `maintenance_timesheet` problem. Locally tested, `maintenance_stock` works properly. I actually solved it on v13 migration, but I forgot porting that update to v12 implementation. As soon as possible I'll make the PR to fix it. Meanwhile I think it's not a problem for this PR approval progress, what do you think?

Open #105

dalonsod:
Update:
- Added `ROADMAP` file.
- Issue with requests button box solved:

imagen

jarroyomorales (approved):
Functional review, LGTM!! :+1:

maintenance_stock/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,65 @@
...
+
+    <record id="hr_equipment_view_form" model="ir.ui.view">
+        <field name="model">maintenance.equipment</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button class="oe_stat_button" type="object" name="action_view_stock_picking_ids"
+                        icon="fa-ticket"
+                        attrs="{'invisible': [('allow_consumptions', '=', False)]}"
+                        groups="stock.group_stock_user">
```
etobella: I prefer to add the permissions on the view. In this case it seems right, but I found issues on real life due to adding the the permissions inside :thinking:

maintenance_stock/views/maintenance_request_views.xml
```diff
@@ -0,0 +1,57 @@
...
+            <xpath expr="//field[@name='category_id']" position="after">
+                <field name="allow_consumptions" />
+            </xpath>
+        </field>
+    </record>
+
+    <record id="equipment_request_view_form" model="ir.ui.view">
+        <field name="model">maintenance.request</field>
+        <field name="inherit_id" ref="base_maintenance.equipment_request_view_form" />
+        <field name="arch" type="xml">
```
etobella, line 31: Idem

maintenance_stock/views/stock_move_line_views.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="stock_move_line_view_search" model="ir.ui.view">
+        <field name="model">stock.move.line</field>
+        <field name="inherit_id" ref="stock.stock_move_line_view_search" />
```
etobella, line 6: Permissions is missing
imagen

etobella (changes requested):
Add permissions on views in order to make sure that everything works fine please :wink:

maintenance_stock/views/maintenance_equipment_views.xml
etobella: It is not failing because you are not showing how many items, are used, but if you add the computed field, that would be a failure probably.

etobella (approved):

maintenance_stock/views/maintenance_equipment_views.xml
dalonsod: No problem, I'll remember it as a good practice. Added also to `search` view

maintenance_stock/views/maintenance_request_views.xml
dalonsod: Done

maintenance_stock/views/stock_move_line_views.xml
dalonsod: I don't really know if this issue should be solved: if a user belogs to "Inventory / User" can access Product Moves in general... but not to every maintenance request, because of record rules. That access error is fired because user, who can access requests, it's not following this one.

Should it be added a record rule to prevent it, or accessing the Product Move is still correct, even if belongs to a request you don't follow to?

etobella: If the permission is added on the view, the error shouldn't happen

dalonsod:
@etobella `maintenance_stock` is still unavailable on weblate. Could you check?

PR #105, opened 2020-09-21
[12.0][FIX] maintenance_timesheet: use button placeholder in request form

dalonsod:
This PR fixes that v12 addon implementation duplicates button box in maintenance request form view (already existing in `base_maintenance` addon). This change is already present in v13, so no cherry pick is needed.

### DIFF
--- a/maintenance_timesheet/views/maintenance_request_views.xml
+++ b/maintenance_timesheet/views/maintenance_request_views.xml
@@ -52,22 +52,16 @@
         </field>
     </record>
 
-    <record id="hr_equipment_request_view_form" model="ir.ui.view">
+    <record id="equipment_request_view_form" model="ir.ui.view">
         <field name="model">maintenance.request</field>
-        <field name="inherit_id" ref="maintenance.hr_equipment_request_view_form" />
-        <field name="priority" eval="110"/>
+        <field name="inherit_id" ref="base_maintenance.equipment_request_view_form" />
         <field name="arch" type="xml">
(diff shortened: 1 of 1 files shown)


marcelsavegnago (approved):

jpcarassato (approved):
OK!

jarroyomorales (approved):

maintenance_timesheet/views/maintenance_request_views.xml
```diff
@@ -55,19 +55,13 @@
     <record id="hr_equipment_request_view_form" model="ir.ui.view">
         <field name="model">maintenance.request</field>
         <field name="inherit_id" ref="maintenance.hr_equipment_request_view_form" />
```
etobella: Maybe you should use the view where the button box is inserted:

https://github.com/OCA/maintenance/blob/11.0/base_maintenance/views/maintenance_request_views.xml#L6

etobella (changes requested):

maintenance_timesheet/views/maintenance_request_views.xml
dalonsod: OK, I'll fix it.

Neither v13 implementation (https://github.com/OCA/maintenance/blob/13.0/maintenance_timesheet/views/maintenance_request_views.xml#L77) uses base_maintenance view inheritance, so it's also incorrect, isn't it?

dalonsod:
@etobella comment attended

etobella (approved):

PR #107, opened 2020-09-23
[12.0][ADD] user see only maintenance team where he is join in

Borruso:
Add ir.rule which it will be able to see only maintenance team where user is join in

### DIFF
--- a/base_maintenance_group/security/maintenance_security.xml
+++ b/base_maintenance_group/security/maintenance_security.xml
@@ -3,22 +3,49 @@
      License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
 
 <odoo>
+    <data noupdate="0">
 
-    <record id="group_maintenance_user" model="res.groups">
-        <field name="name">User</field>
-        <field name="category_id" ref="module_maintenance"/>
-    </record>
+        <record id="group_maintenance_user" model="res.groups">
+            <field name="name">User</field>
+            <field name="category_id" ref="module_maintenance"/>
+        </record>
 
-    <record id="maintenance.group_equipment_manager" model="res.groups">
-        <field name="category_id" ref="module_maintenance"/>
-        <field name="implied_ids" eval="[(4, ref('group_maintenance_user'))]"/>
-    </record>
+        <record id="maintenance.group_equipment_manager" model="res.groups">
+            <field name="category_id" ref="module_maintenance"/>
+            <field name="implied_ids" eval="[(4, ref('group_maintenance_user'))]"/>
+        </record>
 
-    <record id="group_maintenance_manager" model="res.groups">
-        <field name="name">Full Access</field>
-        <field name="category_id" ref="module_maintenance"/>
-        <field name="implied_ids" eval="[(4, ref('maintenance.group_equipment_manager')), (4, ref('group_maintenance_user'))]"/>
-        <field name="users" eval="[(4, ref('base.group_system'))]"/>
-    </record>
+        <record id="group_maintenance_manager" model="res.groups">
+            <field name="name">Full Access</field>
+            <field name="category_id" ref="module_maintenance"/>
+            <field name="implied_ids" eval="[(4, ref('maintenance.group_equipment_manager')), (4, ref('group_maintenance_user'))]"/>
+            <field name="users" eval="[(4, ref('base.group_system'))]"/>
+        </record>
 
+    </data>
+    <data noupdate="1">
+
+        <record id="only_subscribe_maintenance_team" model="ir.rule">
+            <field name="name">Only Subscribe Maintenance Team</field>
+            <field name="model_id" ref="maintenance.model_maintenance_team"/>
+            <field name="groups" eval="[(6, 0, [ref('group_maintenance_user')])]"/>
+            <field name="domain_force">[('member_ids', 'in', [user.id])]</field>
+            <field name="perm_read" eval="True"/>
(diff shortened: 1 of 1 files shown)


base_maintenance_group/security/maintenance_security.xml
```diff
@@ -21,4 +21,15 @@
         <field name="users" eval="[(4, ref('base.group_system'))]"/>
     </record>
 
+    <record model="ir.rule" id="only_subscribe_maintenance_team">
```
astirpe: Please place `id` attribute before `model`, as stated in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#xml-files:


```suggestion
    <record id="only_subscribe_maintenance_team" model="ir.rule">
```

astirpe (changes requested):
Thanks for this PR.
Only a minor request to adopt the conventions as per the OCA guidelines.

jarroyomorales:
Hi @Borruso! I dont see any problem in being able to see other teams in the company. It may be interesting for some reason.

Borruso:
Could I have to make rule as well for version 13.0 ??

astirpe:
> Could I have to make rule as well for version 13.0 ??

Let's see if this PR is accepted a merged first. In case, after merged, you could create a new PR with the porting to V13.

dalonsod:
> Hi @Borruso! I dont see any problem in being able to see other teams in the company. It may be interesting for some reason.

@Borruso maybe this rule could be added within a `<data noupdate="1">` section. And IMO you should create another rule to grant total access for the `Full Access` group.

base_maintenance_group/security/maintenance_security.xml
```diff
@@ -21,4 +21,30 @@
...
+            <field name="perm_write" eval="True"/>
+            <field name="perm_unlink" eval="True"/>
+            <field name="perm_create" eval="True"/>
+        </record>
+
+        <record id="all_maintenance_team" model="ir.rule">
+            <field name="name">All Maintenance Team</field>
+            <field name="model_id" ref="maintenance.model_maintenance_team"/>
+            <field name="groups" eval="[(6, 0, [ref('maintenance.group_equipment_manager')])]"/>
+            <field name="domain_force">[(1, 'in', 1)]</field>
```
dalonsod: This should not work:
```suggestion
            <field name="domain_force">[(1, '=', 1)]</field>
```

dalonsod (changes requested):

dalonsod (approved):
Tested on runbot 👍 
Not blocking, I only suggest a cosmetic change: for better understanding of `maintenance_security.xml` file after these changes: the already existing records should be placed under a `<data noupdate="0">` section

marcelsavegnago (approved):
LGTM

etobella (approved):

dalonsod:
> Not blocking, I only suggest a cosmetic change: for better understanding of `maintenance_security.xml` file after these changes: the already existing records should be placed under a `<data noupdate="0">` section

Maybe I've not explained this well. The suggested change should be keep `<data noupdate="1">` for the recently added rules and include `<data noupdate="0">` for the previous code (the `res.groups` records), in order to make the code more readable. @etobella do you agree?

Borruso:
> > Not blocking, I only suggest a cosmetic change: for better understanding of `maintenance_security.xml` file after these changes: the already existing records should be placed under a `<data noupdate="0">` section
> 
> Maybe I've not explained this well. The suggested change should be keep `<data noupdate="1">` for the recently added rules and include `<data noupdate="0">` for the previous code (the `res.groups` records), in order to make the code more readable. @etobella do you agree?

Sorry, I didn't understand. Should I wait @etobella or not?

dalonsod:
@etobella approved your recent changes, thus I'd wait for him

base_maintenance_group/security/maintenance_security.xml
```diff
@@ -21,4 +21,30 @@
         <field name="users" eval="[(4, ref('base.group_system'))]"/>
     </record>
 
+    <data noupdate="0">
```
etobella: Sorry, noupdate should be 1, as usual on record rules

etobella (changes requested):
Follow @dalonsod comments, It is just cosmetic but easier to understand. Also, I saw the noupdate and it should be 1, not 0 on the rule, as we could make a change there. If you want, you could create the data on a migration script in order to be sure that are created, but make them unactive on the creation as we don't want to break the old logic.

dalonsod (approved):
:+1:

etobella:
If you update the module, you will not get the new rules on your system, maybe you need a migration script in order to add them

etobella:
Here you can check an example:
https://github.com/OCA/community-data-files/blob/12.0/account_payment_unece/migrations/12.0.1.0.0/post-migration.py#L7

etobella:
@Borruso Have you checked?

Borruso:
> @Borruso Have you checked?

@etobella I checked and I understand that even though the rules are in noupdate, no migration script is needed since these rules do not exist in the system, the problem shows up when the rules already existed in the system and I changed, for example, name or domain

etobella (approved):
You are right! Tested locally, when upgrading, the rules are created

marcelsavegnago (approved):

Borruso:
@astirpe could you check if your change request is ok?

astirpe (approved):
Code review ok. Not tested.

PR #114, opened 2020-11-16
[14.0][MIG] base_maintenance

dalonsod:
Standard migration


etobella (approved):

lmiguens-solvos (approved):
LGTM

base_maintenance/views/maintenance_team_views.xml
```diff
@@ -0,0 +1,38 @@
...
+                <group>
+                    <field
+                        name="description"
+                        placeholder="Description of the team's function"
+                    />
+                </group>
+            </xpath>
+        </field>
+    </record>
+    <!-- There is no search view to inherit, so creating it-->
```
jarroyomorales: This search view is already added in Odoo's core so its not needed here. I just realized it was like this also in v13 and v12, so maybe we should do a PR fixing it.

base_maintenance/views/report_maintenance_request.xml
```diff
@@ -106,14 +106,19 @@
...
-        <report
-            id="action_report_maintenance_request"
-            string="Maintenance Request"
-            model="maintenance.request"
-            report_type="qweb-pdf"
-            file="base_maintenance.report_maintenance_request"
-            name="base_maintenance.report_maintenance_request"
-            print_report_name="'Maintenance Request - %s' % (object.name)"
-        />
+        <record id="action_report_maintenance_request" model="ir.actions.report">
```
jarroyomorales, line 109: The report doesnt appear in the request's form

jarroyomorales (approved):

base_maintenance/views/maintenance_team_views.xml
dalonsod: Done!

base_maintenance/views/report_maintenance_request.xml
dalonsod: @jarroyomorales  I think the problem is that report record needs `binding_type` and `binding_model_id` fields. The first one is already fixed, but the second one not yet. On runbot I've clicked on "Add in the print menu" button to achieve it.

In my local environment adding this line in order to permanently fix it:

`<field name="binding_model_id" ref="model_maintenance_request" />`

causes a parse error. And I didn't solve it yet... :(

jarroyomorales: Oh, I tested on runbot and saw the button, thats why I approved the PR. 
Maybe `ref="maintenance.model_maintenance_request"` solves it?

dalonsod: Thank you so much! :D

PR #115, opened 2020-11-30
[IMP] base_maintenance: remove unnecessary view

jarroyomorales:
As said in https://github.com/OCA/maintenance/pull/114#discussion_r532464511

@etobella @dalonsod

### DIFF
--- a/base_maintenance/views/maintenance_team_views.xml
+++ b/base_maintenance/views/maintenance_team_views.xml
@@ -19,15 +19,14 @@
         </field>
     </record>
 
-    <!-- There is no search view to inherit, so creating it-->
     <record id="maintenance_team_view_search" model="ir.ui.view">
         <field name="name">maintenance.team.search</field>
         <field name="model">maintenance.team</field>
+        <field name="inherit_id" ref="maintenance.maintenance_team_view_search"/>
         <field name="arch" type="xml">
-            <search>
-                <field name="name"/>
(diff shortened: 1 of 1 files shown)


jarroyomorales:
Finally I will inherit the original's view and add the name as in newer versions.

base_maintenance/views/maintenance_team_views.xml
```diff
@@ -19,15 +19,14 @@
         </field>
     </record>
 
-    <!-- There is no search view to inherit, so creating it-->
     <record id="maintenance_team_view_search" model="ir.ui.view">
```
etobella, line 22: On 13.0 and 14.0 this view is no longer necessary :smile:

etobella (approved):

base_maintenance/views/maintenance_team_views.xml
jarroyomorales: For v14 it is removed in the migration, but for 13.0 we may have to leave it like this. Others might be inheriting from it and we can cause them some errors.

dalonsod:
Maybe commit message should be modified, because the view still exists, and, I don't sure but a rebase should be needed to pass codecov

jarroyomorales:
@dalonsod Done!

dalonsod (approved):
I don't understand codecov errors... as far as I can see, the PR LGTM

PR #117, opened 2020-12-02
[14.0][MIG] maintenance_equipment_status

dalonsod:
Standard migration + some demo data


dalonsod:
@AdriaGForgeFlow @JordiBForgeFlow could you review? Thanks!

AdriaGForgeFlow (approved):
LGTM :+1:

maintenance_equipment_status/models/maintenance_equipment.py
```diff
@@ -0,0 +1,17 @@
...
+
+class MaintenanceEquipment(models.Model):
+    _inherit = "maintenance.equipment"
+
+    status_id = fields.Many2one(
+        string="Status",
+        comodel_name="maintenance.equipment.status",
+        tracking=True,
+    )
+    status_name = fields.Char(
```
jarroyomorales: Is this field necessary?

dalonsod: `status_name` is used in equipment kanban view

jarroyomorales: Yes, but im not sure if you could do `<t t-raw="record.statusid.name.raw_value" />` or something like this and delete the field

dalonsod: OK, I understand... Locally tested, the module got installed, but a Javascript error is thrown when trying to load kanban view.
The changes made were
`<field name="status_id" />` instead of `<field name="status_name" /> `
and 
`record.status_id.name.raw_value` instead of `record.status_name.raw_value`

jarroyomorales: Could yo try with `record.status_id.value`? I tested it on v13 and it worked

dalonsod: it worked, thanks!
I'll safely remove that field, and then force push. @AdriaGForgeFlow is there any problem if `status_name` related field is removed?

jarroyomorales (approved):

maintenance_equipment_status/models/maintenance_equipment.py
AdriaGForgeFlow: I think there should be no problem removing the field :+1:

etobella (approved):

PR #118, opened 2020-12-03
[13.0][MIG] maintenance_equipment_contract

jarroyomorales:
Standard migration

@etobella @dalonsod


dalonsod:
@jarroyomorales `oca_dependencies.txt` with `contract` line is needed

jarroyomorales:
@dalonsod Green already! :green_apple:

maintenance_equipment_contract/models/maintenance_equipment.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class MaintenanceEquipment(models.Model):
+
+    _inherit = "maintenance.equipment"
+
+    contract_ids = fields.Many2many("contract.contract", string="Contracts")
+    contract_count = fields.Integer(compute="_compute_contract_count",)
+
+    @api.depends("contract_ids")
```
dalonsod, line 14: Since `contract_count` is a non-stored computed field, is actually this decorator applied?

maintenance_equipment_contract/views/maintenance_equipment.xml
```diff
@@ -0,0 +1,25 @@
...
+                    name="action_view_contracts"
+                    class="oe_stat_button"
+                    icon="fa-usd"
+                    type="object"
+                >
+                    <field name="contract_count" widget="statinfo" string="Contracts" />
+                </button>
+            </div>
+        </field>
+    </record>
```
dalonsod, line 24: Suggested improvement (not blocking): add a `search` view with filters "Has contracts" / "Has no contracts"

dalonsod (changes requested):
See comments, only code review

maintenance_equipment_contract/views/maintenance_equipment.xml
jarroyomorales: Added them

maintenance_equipment_contract/models/maintenance_equipment.py
jarroyomorales: If I remove it the test will fail. It can probably be solved with a `.refresh()` but I think it does not hurt to keep the depends and may give some unexpected behaviour in some cases

etobella (approved):

etobella:
@dalonsod Is it right then?

dalonsod (approved):
Tested on runbot 👍

PR #125, opened 2021-01-04
[10.0][BACKPORT] maintenance_equipment_hierarchy

baimont:
backport of maintenance_equipment_hierarchy from 11.0 to 10.0


jarroyomorales:
Red travis, but not sure if its because of this module.. Could you check?

baimont:
> Red travis, but not sure if its because of this module.. Could you check?

Hey @jarroyomorales, seems to be linked with postgres server :

```
createdb: could not connect to database template1: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

I don't see how my pr would cause that problem.

apineux (approved):

etobella:
@baimont I made a PR to solve your issue, as soon as it is merged, you could rebase and the force-push and everything should be fixed :smile:

baimont:
> @baimont I made a PR to solve your issue, as soon as it is merged, you could rebase and the force-push and everything should be fixed smile

thanks @etobella  :-)

maintenance_equipment_hierarchy/models/maintenance_equipment.py
```diff
@@ -0,0 +1,69 @@
...
+    def preview_child_list(self):
+        context = self.env.context.copy()
+        return {
+            'name': 'Child equipment of %s' % self.name,
+            'type': 'ir.actions.act_window',
+            'res_model': 'maintenance.equipment',
+            'res_id': self.id,
+            'view_mode': 'list,form',
+            'context': context.update({
+                'default_parent_id': self.id,
```
etobella: The context is not working properly.

etobella (changes requested):
The UTF-8 headers are missing on all py files, isn't it?

maintenance_equipment_hierarchy/models/maintenance_equipment.py
```diff
@@ -0,0 +1,72 @@
...
+            'parent_id_editable': False,
+        })
+        return {
+            'name': 'Child equipment of %s' % self.name,
+            'type': 'ir.actions.act_window',
+            'res_model': 'maintenance.equipment',
+            'res_id': self.id,
+            'view_mode': 'list,form',
+            'context': context,
+            'domain': [('id', 'in', self.child_ids.ids)],
```
etobella: Change it by 'domain': [('parent_id', '=', self.id)]

This way, if you create a new children from this view it will work properly.
Right now, when you create a child, it is not shown into the tree until you press again the button `Children`

etobella:
Can you join your commits and then we can proceed with merge?

PR #126, opened 2021-01-08
[FIX] travis

etobella:
This is probably the issue on #125


etobella:
Fast tracking as Travis is not working /ocabot merge nobump

PR #127, opened 2021-01-19
[14.0][MIG] maintenance_timesheet

dalonsod:
Standard migration


dalonsod:
@marcelsavegnago @jarroyomorales could you review? Thanks!

maintenance_timesheet/models/maintenance_request.py
```diff
@@ -0,0 +1,57 @@
...
+    def action_view_timesheet_ids(self):
+        """
+        Access to the current timesheets for this maintenance request
+        The view will be restricted to the current request and only HR managers
+        could create timesheets for every employee
+        """
+        self.ensure_one()
+        action = (
+            self.env.ref("maintenance_timesheet.timesheet_action_from_request")
+            .with_user(self.env.ref("base.user_admin"))
```
etobella: Why this-???

dalonsod: When testing demo user was not allowed to access to actions model, maybe is a new v14 feature. I didn't find another solution to achieve the desired behavior

maintenance_timesheet/models/maintenance_request.py
```diff
@@ -0,0 +1,57 @@
...
+            .read()[0]
```
dalonsod: @etobella I've found that this change seems to work:

```suggestion
            self.env["ir.actions.act_window"]._for_xml_id("maintenance_timesheet.timesheet_action_from_request")
```

I've checked https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 and this possible migration task is not listed. @pedrobaeza is that change correct when migrating to v14?

dalonsod: > 
> 
> @etobella I've found that this change seems to work:
> 
> I've checked https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 and this possible migration task is not listed. @pedrobaeza is that change correct when migrating to v14?

@pedrobaeza @etobella could you check this? If the proposed approach is the right one, I'll update the PR. Thanks!

pedrobaeza: I don't see this as a general change for 14.0, but specific business logic one, or am I missing something?

dalonsod: I've found that demo user with his default permissions can't apparently access` ir.actions.act_window` model in readonly mode, and that didn't happen in v13, as far as I know. So, raw accesses (such as `read` method in order to change domain or context of an action) to this model will fail. Then, IMO during an addon migration I think this kind of code could be present sometimes, and I asked for the best practice you should recommend.

Anyway if `_for_xml_id()` does the trick, there's no problem in this case for me, and I'll use it.

dalonsod:
ping @marcelsavegnago @jarroyomorales

etobella (approved):

patrickrwilson (approved):
Tested and works as intended. LGTM

etobella:
Travis failed @dalonsod Can you rebase in order to check why?

dalonsod:
> Travis failed @dalonsod Can you rebase in order to check why?

@etobella done

PR #134, opened 2021-07-05
[12.0][FIX] maintenance_plan: migration script wrong syntax

dalonsod:
With current syntax an "ambiguous column" error is fired, in `WHERE` clause I think. Also, in `SET` clause the table alias points to the wrong table.

cc @AdriaGForgeFlow @AaronHForgeFlow


AaronHForgeFlow:
Doing the set using the maintenance_equipment table was done on purpose because the column is empty in maintenance_plan table, we are filling in.

Ok to add the reference in the WHERE to be `meq.period`

etobella (approved):

PR #136, opened 2021-07-05
[12.0][FIX] maintenance_plan: preventive request generation with the same company of its equipment

dalonsod:
With this change, in a multi-company environment a request is created with the proper company, rather than OdooBot one.

cc @AdriaGForgeFlow @etobella

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -50,6 +50,7 @@ class MaintenanceEquipment(models.Model):
 
         return {
             'name': name,
+            'company_id': self.company_id.id,
             'request_date': next_maintenance_date,
             'schedule_date': next_maintenance_date,
             'category_id': self.category_id.id,
@@ -104,7 +105,7 @@ class MaintenanceEquipment(models.Model):
         for plan in self.env['maintenance.plan'].search(
                 [('interval', '>', 0)]):
             equipment = plan.equipment_id
-            equipment._create_new_request(plan)
+            equipment.sudo()._create_new_request(plan)
 
     @api.depends('maintenance_plan_ids.next_maintenance_date',
                  'maintenance_ids.request_date')


maintenance_plan/models/maintenance_equipment.py
```diff
@@ -104,7 +105,7 @@ def _cron_generate_requests(self):
         for plan in self.env['maintenance.plan'].search(
                 [('interval', '>', 0)]):
             equipment = plan.equipment_id
-            equipment._create_new_request(plan)
+            equipment.sudo()._create_new_request(plan)
```
etobella, line 108: Maybe a `with_context(force_company=self.company_id.id)` is enough. Can you try it?

dalonsod: Not working, I think because `self.company_id.id` points to OdooBot company, not equipment one. The requests are created, but in the wrong company

etobella: self.company_id points to the equipment company, self.env.company points to User company, so I don't understand it...

dalonsod: Tested with `with_context(force_company=equipment.company_id.id)` and it didn't work. I'm not an expert in cross-company configurations, but `force_company` works in v12?

etobella: I found the problem... maybe you should need to add `"company_id": self.company_id.id` on `_prepare_request_from_plan` that you can find in maintenance.equipment

dalonsod: I don't understand, that line is already added in this PR. The current implementation (that line + `sudo()` in cron in order to prevent OdooBot company access issues) it's working...

etobella: Sorry, I was checking locally, I think the problem might be related to the rule, maybe OdooBot has no access to this new company? In this case, the sudo makes sense.

etobella (approved):

maintenance_plan/models/maintenance_equipment.py
dalonsod: > I think the problem might be related to the rule, maybe OdooBot has no access to this new company?

That's right: I've got an installation with a similar issue, and I've solved it adding every existing company to OdooBot, but IMO this is a better and permanent solution

PR #138, opened 2021-07-14
[13.0][FIX TRAD] maintenance_project - fr translation was not working

baimont:


etobella:
The translations must be managed on weblate. I created the file for you:

https://translation.odoo-community.org/projects/maintenance-13-0/maintenance-13-0-maintenance_project/fr/

baimont:
> The translations must be managed on weblate. I created the file for you:
> 
> https://translation.odoo-community.org/projects/maintenance-13-0/maintenance-13-0-maintenance_project/fr/

Hi and thanks for your reply.
I have the exact same problem here https://github.com/OCA/delivery-carrier/pull/394 and here https://github.com/OCA/maintenance/pull/137

I went on weblate and tried to create a new translation for fr, as you created for maintenance_project. It seems I don't have the rights. Why is that?

EDIT : I think I found alone : I added english and french in the settings of my weblate account. And now I can create a new translation for these languages.

I still have a question : if we merge this pr, won't weblate be updated? fr_FR translation is not recognized by odoo13 and is therefore useless. If weblate would be updated it is a better solution to merge this pr so that we don't have to translate towards french a second time.

etobella:
You should cancel this PR and use weblate. That would be the right approach

PR #154, opened 2021-09-08
ci: run tests with GitHub actions

etobella:
Applies the same logic from https://github.com/OCA/mis-builder/pull/365
:crossed_fingers:


etobella:
@JordiBForgeFlow I made a relicense on a module that you made. Can you check it?

LoisRForgeFlow (approved):
LGTM, including the re-license of the ForgeFlow module :+1:

etobella:
@sbidoul runbot is not launching. Something else is necessary or in this case there will be no more runbot?

sbidoul:
@etobella you need to keep `.travis.yml` until runbot is updated.

etobella:
But if the file exists, travis will be executed. Isn't it?

sbidoul:
> But if the file exists, travis will be executed. Isn't it?

Yes, unfortunately...

etobella:
Well runbot did not launch here also: https://github.com/OCA/maintenance/pull/155
Also, with the travis file restored (the first one prevents travis execution), it is not launched :sob: 
Any ideas @sbidoul ?

sbidoul:
Maybe trying opening a new PR that does not remove .travis.yml ?

etobella:
It has been restored and travis was executed

etobella:
I will create a new PR. Let's see what happens ...

PR #153, opened 2021-09-06
[13.0][FIX] base_maintenance: maintenance team view fixes

LoisRForgeFlow:
* search view already in core. Not need to forward port to v14, fixed there during migration.
* show user_id and member_ids in tree view. Since the tree view is editable when accessing in the configuration menu, the new field need to be showed also in the tree view.

@ForgeFlow

PR #155, opened 2021-09-09
[14.0][FIX] base_maintenance: show user_id and member_ids in tree view.

LoisRForgeFlow:
Since the tree view is editable when accessing in the configuration
menu, the new field need to be showed also in the tree view.

Forward port of https://github.com/OCA/maintenance/pull/153.

@ForgeFlow


etobella (approved):

etobella:
I think something happened with ocabot. Let's try again:
/ocabot merge minor

Issue #161, opened 2021-09-29
[BUG] maintenance_plan: next maintenance date do not consider recently done maintenance requests


LoisRForgeFlow:
This affect all versions of this module but I will use v13 to show the problem. 

Imagine you have a maintenance plan, say annually, and the next maintenance date is 2021-10-01 and you have already the maintenance request in place to be completed. At this point the next maintenance date showed in the plan is 2021-10-01, which is correct. However if you finish the maintenance request some days before the deadline, for instance on 2021-09-29, the next maintenance date will continue to be 2021-10-01 which is not correct anymore.

![image](https://user-images.githubusercontent.com/23449160/135279472-199163e2-db68-414f-9b3c-0393cd712491.png)

![image](https://user-images.githubusercontent.com/23449160/135279561-76f3dc54-88c9-44da-a90c-a1d8a5579cc0.png)

[This](https://github.com/OCA/maintenance/blob/14.0/maintenance_plan/models/maintenance_plan.py#L121) is the problematic method:

```python
@api.depends(
        "interval",
        "interval_step",
        "maintenance_kind_id",
        "equipment_id.maintenance_ids.request_date",
        "equipment_id.maintenance_ids.close_date",
        "equipment_id.maintenance_ids.maintenance_kind_id",
    )
    def _compute_next_maintenance(self):

        for plan in self.filtered(lambda x: x.interval > 0):

            interval_timedelta = get_relativedelta(plan.interval, plan.interval_step)

            next_maintenance_todo = self.env["maintenance.request"].search(
                [
                    ("equipment_id", "=", plan.equipment_id.id),
                    ("maintenance_type", "=", "preventive"),
                    ("maintenance_kind_id", "=", plan.maintenance_kind_id.id),
                    ("maintenance_plan_id", "=", plan.id),
                    ("stage_id.done", "!=", True),
                    ("close_date", "=", False),
                ],
                order="request_date asc",
                limit=1,
            )

            if next_maintenance_todo:
                plan.next_maintenance_date = next_maintenance_todo.request_date
            else:
                next_date = plan.start_maintenance_date
                while next_date < fields.Date.today():
                    next_date = next_date + interval_timedelta
                plan.next_maintenance_date = next_date
```

We have 2 options here, consider the next date as:
A) 2022-10-01, this respects the maintenance plan.
B) 2022-09-29, this is one year from the last maintenance request completion.

What do you think would be the best option to fix this bug?

cc @etobella @dalonsod

[...]

PR #162, opened 2021-09-29
[FIX] maintenance_plan: Remake the computation of next date

etobella:
Solves #161

### DIFF
--- a/maintenance_plan/models/maintenance_plan.py
+++ b/maintenance_plan/models/maintenance_plan.py
@@ -112,37 +112,42 @@ class MaintenancePlan(models.Model):
     @api.depends(
         "interval",
         "interval_step",
-        "maintenance_kind_id",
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
Some adjustments here https://github.com/tegin/maintenance/pull/1.

LoisRForgeFlow (approved):
LGTM now :+1:

maintenance_plan/models/maintenance_plan.py
```diff
@@ -112,37 +112,41 @@ def _compute_maintenance_count(self):
     @api.depends(
         "interval",
         "interval_step",
```
LoisRForgeFlow, line 114: I think `start_maintenance_date` should also be in the depends.

mariadforgeflow (approved):

PR #170, opened 2021-10-06
[FIX] maintenance_equipment_contract: fix security for viewing contra…

albariera:
…cts from equipment

A maintenance manager user could not view contracts if he had not accounting permissions, with this PR this problem gets fixed


etobella (approved):

LoisRForgeFlow (approved):
LGTM

Issue #159, opened 2021-09-28
[IMP] maintenance_sequence: code, partner_ref and serial_no


etobella:
Reviewing odoo code and OCA code I found that fields `code` and `serial_no` should fulfill the same logic. 

Serial number is a field that should be used to add our internal serial number, as it is unique in our database. On the other side, partner_ref should be used to write the serial number of the vendor/manufacturer. Both fields are searchable using `name`:

https://github.com/odoo/odoo/blob/master/addons/maintenance/views/maintenance_views.xml#L507-L509

So, code and serial_no are internal identifier. I think the sequence should fulfill serial_no. The biggest problem is how to manage the migration, as it could be complicated to manage (Maybe on the migration from 14 to 15?)

WDYT?

@dalonsod @LoisRForgeFlow

LoisRForgeFlow:
That makes total sense, thanks for the heads-ups. For me it could even be possible to propose the change in v14 and even v13 if migration scripts are provided.

My 2 cents.

dalonsod:
@etobella inspecting the code I've found that `code` isn't actually unique, so you can assign the same `code` to more than one equipment. I think it's an issue rather than a feature... then, if a migration script e.g. should move current codes to `serial_no`, when codes are filled and serial numbers not, an error could happen.

etobella:
Yes, it is not unique, but it will probably be unless you make the sequence wrongly

etobella:
Also, that's the work of a great migration script :wink:

dalonsod:
> 
> 
> Yes, it is not unique, but it will probably be unless you make the sequence wrongly

Or you manually fill the same `code` twice, bypassing sequence automation.

PR #174, opened 2021-10-22
[15.0][MIG] maintenance_equipment_sequence

BernatPForgeFlow:


BernatPForgeFlow:
@ForgeFlow 
@LoisRForgeFlow 
@AaronHForgeFlow

dalonsod:
@etobella maybe it's time to resume #159 , what do you think?

maintenance_equipment_sequence/models/maintenance.py
```diff
@@ -0,0 +1,129 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class MaintenanceEquipmentCategory(models.Model):
+    _inherit = "maintenance.equipment.category"
+
+    sequence_prefix = fields.Char(
+        name="Sequence Prefix",
```
MiquelRForgeFlow:
```suggestion
        string="Sequence Prefix",
```

maintenance_equipment_sequence/tests/test_maintenance_equipment_sequence.py
```diff
@@ -0,0 +1,61 @@
...
+        equipment_1 = self.maintenance_equipment_obj.create(
+            {"name": "Laptop 1", "category_id": categ_2.id}
+        )
+        categ_2._compute_equipment_code()
+        self.assertEqual(equipment_1.code, "TTC0001")
+        # Set code manually
+        equipment_1.write({"code": "TTC0023"})
+        self.assertEqual(equipment_1.code, "TTC0023")
+        # Remove code and be automatically set to sequence next value
+        equipment_1.write({"code": False})
```
MiquelRForgeFlow:
```suggestion
        self.assertEqual(equipment_1.serial_no, "TTC0001")
        # Set code manually
        equipment_1.write({"serial_no": "TTC0023"})
        self.assertEqual(equipment_1.serial_no, "TTC0023")
        # Remove code and be automatically set to sequence next value
        equipment_1.write({"serial_no": False})
```

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Column `code` of table `maintenance_equipment` has been modified to not have duplicates
+    and then has been copied to `serial_no` column because they have the same function and
+    `serial_no` is the one in the core.
+    Afterwards, column 'code' has been deleted.
+    """
+
+    if not version:
+        return
+
```
MiquelRForgeFlow, line 40: add
```suggestion

    cr.execute(
        """
        UPDATE maintenance_equipment
        SET code = serial_no
        WHERE code IS NULL;"""
    )

```
in order to avoid unexpected rare case:
```
coming with:

a: id=1, code=NULL, serial_nu=5
b: id=2, code=5, serial_nu=XX

and you end with:

a: id=1, serial_nu=5
b: id=2, serial_nu=5
```

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,47 @@
...
+    cr.execute(
+        """
+        UPDATE maintenance_equipment
+        SET code = serial_no
+        WHERE code IS NULL;"""
```
etobella: WHERE serial_no is NOT NULL

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Column `code` of table `maintenance_equipment` has been modified to not have duplicates
+    and then has been copied to `serial_no` column because they have the same function and
+    `serial_no` is the one in the core.
+    Afterwards, column 'code' has been deleted.
+    """
+
+    if not version:
+        return
+
```
etobella, line 40: It would be great to create two new legacy columns with the old values (code and serial_no). This way if an implementation worked with serials but wanted to change the code to serial no by default and move the old serial no to a new field , they can execute a script later. Would be great to have an example of this SQL script on the comment of the migration.

In order to be able to pass this changes to older versions. Check if  the code column exists. In this case, the migration has already been done.

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
MiquelRForgeFlow: no

maintenance_equipment_sequence/views/maintenance_views.xml
```diff
@@ -0,0 +1,63 @@
...
+                </group>
+            </group>
+        </field>
+    </record>
+    <!-- equiment : views -->
+    <record id="hr_equipment_view_form" model="ir.ui.view">
+        <field name="name">equipment.form</field>
+        <field name="model">maintenance.equipment</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_view_form" />
+        <field name="arch" type="xml">
```
etobella, line 31: Put a 99 priority, as other modules could have problem with the duplication of the field.

Also make the original serial_no invisible

maintenance_equipment_sequence/views/maintenance_views.xml
```diff
@@ -0,0 +1,63 @@
...
+            <xpath expr="//div[hasclass('oe_title')]" position="before">
+                <div class="row">
+                    <div class="col-xs-2" style="padding-left:1.0em">
+                        <label for="serial_no" class="oe_edit_only" string="Code" />
+                        <h3>
+                            <field name="serial_no" />
+                        </h3>
+                    </div>
+                </div>
+            </xpath>
```
etobella:
```suggestion
            <xpath expr="//div[hasclass('oe_title')]/*[1]" position="before">
                <label for="code" class="oe_edit_only" />
                <h3>
                    <field name="code" />
                </h3>
            </xpath>
```

MiquelRForgeFlow: but `code` field has been deleted.

etobella: I ment serial_no :laughing:

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
etobella: serial_no should be better than code. So we should need to keep if needed. I a customer has a wrong implementation can execute a SQL query in order to get them, as I proposed to keep the old data

MiquelRForgeFlow: ok :+1:

maintenance_equipment_sequence/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        WITH cte AS
+        (
+            SELECT ROW_NUMBER() OVER(PARTITION BY code ORDER BY code ) AS rno, code
+            FROM maintenance_equipment
+        )
+        UPDATE maintenance_equipment me
+        SET code = me.code || '_' || me.id
+        FROM cte
+        WHERE cte.rno>1 AND cte.code = me.code"""
```
MiquelRForgeFlow:
```suggestion
            SELECT ROW_NUMBER() OVER(PARTITION BY code ORDER BY code,id ) AS rno, code, id
            FROM maintenance_equipment
        )
        UPDATE maintenance_equipment me
        SET code = me.code || '_' || cte.rno
        FROM cte
        WHERE cte.rno>1 AND cte.id = me.id"""
```

AaronHForgeFlow (approved):
Functional review (including migration scripts) :+1:

MiquelRForgeFlow (approved):

etobella (approved):

PR #178, opened 2021-11-02
[11.0][FIX] maintenance_plan_activity: pass the res_model in the context so it is used in the team domain

AaronHForgeFlow:
Currently activities are being created for a random team because this domain is not taken the model: https://github.com/OCA/social/blob/11.0/mail_activity_team/models/mail_activity.py#L13

The only relevant change is passing the context all the rest are cosmetics to pass the flake8 checks

@ForgeFlow


LoisRForgeFlow (approved):
LGTM

etobella (approved):

PR #185, opened 2021-11-15
[11.0][FIX] base_maintenance_group: bug in security groups

MiquelRForgeFlow:
The 'users' parameter is res.users and 'base.group_system' is res.groups. Two distinct models!!! It should be base.user_root instead.


MiquelRForgeFlow:
Was an error introduced in https://github.com/OCA/maintenance/commit/1486b56d2271027b930521fd7a9634fd07b0f6ac.

etobella (approved):

MiquelRForgeFlow:
Waiting https://github.com/OCA/maintenance/pull/188.

PR #188, opened 2021-11-16
[11.0][FIX] maintenance_plan_activity: avoid error in indeterministic test

MiquelRForgeFlow:
Selection_052


AaronHForgeFlow (approved):

AaronHForgeFlow:
@ForgeFlow 
@etobella

MiquelRForgeFlow:
I don't know if I should forward port it.

etobella:
This has no sense for me, you are saying that it can be between 2 and 4??? Something different might be happening, but your solution has no sense without some more explanation

etobella:
Executing again the tests are working correctly, IMO this has no sense and the problem came from another site, so I close PR

PR #197, opened 2021-12-20
[13.0][FIX] maintenance_equipment_hierarchy: fix view archived fields and add trasnlatable field

albariera:


maintenance_equipment_hierarchy/models/maintenance_equipment.py
```diff
@@ -53,7 +53,7 @@ def _compute_complete_name(self):
 
     def preview_child_list(self):
         return {
-            "name": "Child equipment of %s" % self.name,
+            "name": _("Child equipment of %s" % self.name),
```
etobella: `_("Child equipment of %s") % self.name,`

etobella (changes requested):
You can do it in a single commit

maintenance_equipment_hierarchy/models/maintenance_equipment.py
albariera: Done :wink:

PR #205, opened 2022-01-12
[15.0][MIG] maintenance_equipment_status

BernatPForgeFlow:


maintenance_equipment_status/data/demo_maintenance_equipment_status.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
cubells: Why do you change this to noupdate=0 ??

cubells (changes requested):

maintenance_equipment_status/data/demo_maintenance_equipment_status.xml
JuanjoA: I think so

BernatPForgeFlow: It should not be changed. Solved!

Issue #206, opened 2022-01-13
[14.0] maintenance_plan: improvable UI design


pedrocs-exo:
The maintenance plan tree view inside the equipment form could use shorter labels. This would make it easier for the user to read and also prevent overlapping the form boundaries.

I'll soon make a PR for turning this:
![image](https://user-images.githubusercontent.com/8116385/149324671-f20a773f-ef6f-48a7-9f24-09989d9431dc.png)

into this:
![image](https://user-images.githubusercontent.com/8116385/149324712-bfdfbcca-9763-49d1-80b0-4f7b4ef583cc.png)

athifbinu:
hello to improve the project ui and style and in css

PR #207, opened 2022-01-13
[14.0][IMP] maintenance_plan: improved embedded tree view labels

pedrocs-exo:
This PR addresses the problems reported on #206


etobella (approved):
Your fixes proably work on your screen, but the same error might happen with a smaller screen :cry: However, I am approving it as it has sense to make titles smaller.

Have you tryied to install web_responsive and see what happens?

pedrocs-exo:
Hi @etobella 
I usually use muk_web_theme. With web_responsive, the sheet's right termination is clean as you can see on the picture:
imagem

But as you said, it still makes sense to shorten the labels. IMHO a few columns could be removed from the tree view and displayed on the form view only.

etobella:
Sure! Can you make a proposal?

dreispt (approved):

PR #209, opened 2022-01-17
[13.0][FIX] maintenance_plan: add rule migration

MiquelRForgeFlow:
Oversight of migration to v13.

### DIFF
--- a/maintenance_plan/migrations/13.0.1.2.4/post-migration.py
+++ b/maintenance_plan/migrations/13.0.1.2.4/post-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2020 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# pylint: disable=C7902
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "maintenance_plan", "migrations/13.0.1.2.0/noupdate_changes.xml"
+    )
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):

etobella (approved):
Seem right. Just one question. It might have sense to add this on a new version change, just for people that already did the migration. WDYT?

etobella (approved):

MiquelRForgeFlow:
Yeah, makes sense. It's done.

maintenance_plan/migrations/13.0.1.2.4/post-migration.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2020 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# pylint: disable=C7902
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr, "maintenance_plan", "migrations/13.0.1.2.0/noupdate_changes.xml"
```
LoisRForgeFlow, line 10: You need to change also the filename here!

LoisRForgeFlow (commented):
This fails in any update with latest code

maintenance_plan/migrations/13.0.1.2.4/post-migration.py
MiquelRForgeFlow: https://github.com/OCA/maintenance/pull/210

PR #202, opened 2021-12-29
[11.0][FIX] maintenance_plan: fix new request date 

mariadforgeflow:
- Fix new request date when plan start date is later.
- Make Request Date visible in tree view.

@LoisRForgeFlow

PR #213, opened 2022-01-24
[12.0][FIX] maintenance_plan: homogenize versions

mariadforgeflow:
Apply same logic as in v11 (#202)


etobella:
But the changes are different

PR #214, opened 2022-01-24
[13.0][FIX] maintenance_plan: homogenize versions

mariadforgeflow:
Apply same logic as in v11 (#202)

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -93,7 +93,8 @@ class MaintenanceEquipment(models.Model):
         )
         # We check maintenance request already created and create until
         # planning horizon is met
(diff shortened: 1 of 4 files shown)


etobella (approved):

marcelsavegnago (approved):

mariadforgeflow:
Hi@etobella! Could you merge this and https://github.com/OCA/maintenance/pull/213 please? Thanks!!

mariadforgeflow:
@etobella  I haven't changed anything in the test that fails. It is failing because there is a test with a monthly plan that starts in today_date and next month is February that has 28 days. This doesn't fail usually and it is set like that in all versions. I don't know if you prefer to wait till tomorrow that won't fail.

mariadforgeflow:
@etobella hi! could you try to merge this again? Thanks:)

PR #208, opened 2022-01-13
[14.0][ADD] maintenance_equipment_image

pedrocs-exo:
A module to add images to equipments

PR #215, opened 2022-01-26
[ADD][12.0] maintenance_equipment_image: new module

cubells:
This module adds image to maintenance equipment. Backported from #208 .


rlizana (approved):
LGTM :+1:

antonio-trey (approved):

etobella:
@cubells Can you take the original module from #208  and make a backport?

cubells:
@etobella it does not work as @pedrocasi did it in #208 because in version 12.0 image mixin does not exist.

So I decided to do it in this version the same way is done in similar models.

etobella (changes requested):
Anyway, you are maintaing the same idea and so on, I think you should keep the original commit and author.

Also, you took the same readmes (nothing was changed there), an extra reason to say that this is a Backport.

cubells:
I don’t agree about who first came up with the idea to put a picture on the equipments, whether @pedrocasi or I.

I think I. I see the #208 later and more so when the code is all rewritten. But I don’t want to argue.

If you ask me, I'll fetch the commits of @pedrocasi .

cubells:
@etobella done!

maintenance_equipment_image/__manifest__.py
```diff
@@ -5,15 +5,15 @@
...
     "summary": """Adds images to equipment.""",
     "category": "Manufacturing/Maintenance",
-    "version": "14.0.1.0.0",
+    "version": "12.0.1.0.0",
     "license": "AGPL-3",
-    "author": "Exo Software," "Odoo Community Association (OCA)",
-    "maintainers": ["pedrocasi"],
+    "author": "Exo Software,"
+              "Odoo Community Association (OCA)",
+    "maintainers": ["cubells"],
```
etobella: Keep pedrocasi please

etobella (commented):
Commit history is somehow weird. For example

https://github.com/OCA/maintenance/pull/215/commits/d2574da8ed6dc99ff55dae0ce08934050adb7d85 
is full of <<<<<<<<< HEAD

maintenance_equipment_image/__manifest__.py
cubells: @etobella done!

cubells:
@etobella can you recheck this?

thanks

etobella (approved):

PR #217, opened 2022-03-15
[14.0][ADD] maintenance_product: New adon.

victoralmau:
This module links *Maintenance Equipments* and *Products*, easing creation in this ways

Please @pedrobaeza and @CarlosRoca13  can you review it?

@Tecnativa TT33410


maintenance_product/models/maintenance_equipment.py
```diff
@@ -0,0 +1,42 @@
...
+        domain="[('categ_id', '=', product_category_id)]",
+    )
+    product_category_id = fields.Many2one(
+        comodel_name="product.category", related="category_id.product_category_id"
+    )
+
+    @api.onchange("product_id")
+    def _onchange_product_id(self):
+        """If product is set, equipment name, seller, seller ref and cost defaults
+        to product ones"""
```
pedrobaeza: Proper PEP257 docstring is:

```suggestion
        """If product is set, equipment name, seller, seller ref and cost defaults
        to product ones.
        """
```

maintenance_product/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module links *Maintenance Equipments* and *Products*, easing creation in this ways
```
pedrobaeza:
```suggestion
This module links *Maintenance Equipments* and *Products*, easing creation of new equipments.
```

maintenance_product/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+Equipment Categories:
+* Go to *Maintenance > Configuration > Equipment Categories* and create a new one (or modify an existing one) filling in the new field *Product Category*.
```
pedrobaeza, line 3: Incorrect RST syntax: Leave one empty line in between (the same for the rest).

maintenance_product/tests/test_maintenance_product.py
```diff
@@ -0,0 +1,40 @@
...
+        cls.product.product_tmpl_id.categ_id = cls.product_category
+
+    def test_maintenance_equipment_category(self):
+        self.category.product_category_id = self.product_category
+        self.category._onchange_product_category_id()
+        self.assertEqual(self.category.name, "test-product-category")
+
+    def test_maintenance_equipment(self):
+        self.equipment.product_id = self.product
+        self.equipment._onchange_product_id()
```
pedrobaeza: Use here `Form` instead (and in the rest of onchanges)

maintenance_product/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Víctor Martínez
+  * Pedro M. Baeza
```
pedrobaeza, line 6: Give credits to César

pedrobaeza (changes requested):

pedrobaeza (approved):

etobella:
If you want to ease the creation, wouldn't have more sense to create the equipment from a purchase?

victoralmau:
> If youbwant tonease the creation, wouldn't have more sense to create the equipment from a purchase?

Perhaps what you indicate would be a different new addon, in this case, without the need for any purchase, you want to link an equipment to a product.

etobella:
I understand the goal of your module, i just asked if it had sense my comment. In my case, we buy equipments but sometimes the product is the same for several versions, as they share the same functionality. Price, supplier and all extra fields are taken from the purchase order. Also equipment is generated on confirmation, as you may need the record before it is received(preparation of spaces...)
I understand this is much more complex. I asked if it had sense to join both ideas somehow

victoralmau:
Ok, thanks for the explanation. Yes, I think it would be a useful use case but I think it should be a new addon called: `maintenance_purchase` that has this as a dependency to be able to create the equipment from the purchase order. What do you think @pedrobaeza ?

pedrobaeza:
Yeah, it's not a bad flow, but we need to ask our customer if they follow such flow.

pedrobaeza:
@etobella we are keeping this flow for now. The other module may be added in the future.

Let's merge it:

/ocabot merge nobump

PR #218, opened 2022-03-30
[14.0][ADD] maintenance_account: New addon.

victoralmau:
Related to: https://github.com/OCA/maintenance/pull/217#issuecomment-1068935212

This module automatically creates the equipment when validating the purchase invoices.

Please @pedrobaeza and @CarlosRoca13  can you review it?

@Tecnativa TT33410


CarlosRoca13 (approved):
Functional and code review :+1:

pedrobaeza:
@etobella finally it's here

hitrosol:
Dear All, I was thinking a year ago about the relation between Equipment in Maintenance Module with Product and Asset Management in Accounting. 

IMHO, equipment can be a Fixed Asset which need to be maintained the Depreciation Schedule in the Assets Management. 
Equipment also need to be treated uniquely, because user will track equipment and put the Fixed Asset Number on it.
In my current practice, I set the product as storable and traceability By Unique Serial Number.  In Asset Management, I set the Asset Profile and check the Create an asset by product item. Therefore, if we purchase more than 1 equipments, i.e: 3 units, and we set the Asset profile in the Vendor Bill, after validate the Vendor Bill the 3 Asset will be created in the Asset Management Module.

In the Inventory, when logistic receive the product, then s/he can put the Fixed Asset Number on it before handed over to the user. The Equipment in maintenance also need to be updated manually with the information from the logistic. But, currently there is no relation between Product received in Inventory with the Maintenance. 

I was thinking, since the equipment is uniquely therefore the relation is supposed to be put in the LOT / Serial Number and Product in Maintenance Module.

In this PR, the new field, Equipment Category is created and will be live side by side with field Asset Profile. Maybe we have different use case but I think having new additional one will make the form too crowded.

pedrobaeza:
For linking assets with maintenance equipments, you have #1

maintenance_account/models/maintenance_equipment.py
```diff
@@ -0,0 +1,18 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class MaintenanceEquipment(models.Model):
+    _inherit = "maintenance.equipment"
+
+    move_id = fields.Many2one(
+        comodel_name="account.move",
+        string="Move",
```
pedrobaeza, line 12: This one can be related from `move_line_id`

maintenance_account/views/account_move_view.xml
```diff
@@ -0,0 +1,46 @@
...
+                </button>
+            </xpath>
+            <xpath
+                expr="//field[@name='invoice_line_ids']/tree/field[@name='quantity']"
+                position="before"
+            >
+                <field
+                    name="equipment_category_id"
+                    attrs="{'column_invisible': [('parent.move_type', '!=', 'in_invoice')]}"
+                    optional="show"
```
pedrobaeza: Better to be hidden by default, as it's not the usual thing you need to do

marcelsavegnago:
ping @douglascstd

pedrobaeza (approved):

etobella:
In my experience, equipments might be created before the invoice is received. I think it might be better to use purchase :thinking:

etobella:
Let's keep this and we might decide later if it is a good idea to propose maintenance_purchase :thinking: 

/ocabot merge nobump

PR #230, opened 2022-06-06
[IMP] maintenance_equipment_tags: add tags to equipment requests

cubells:
Improving module by adding tags to maintenance requests.
Tags are propagated from equipment to request.


rlizana (approved):
:+1:  LGTM

etobella (commented):
I don't really like the change, as you are using the same tags for equipments and requests and it has no sense in my opinion... I would prefer to have a different set of tags with no relation IMO

cubells:
@etobella thanks for the revision.

Now, they have different tags as you requested.

cubells:
@etobella Travis error is not related to module.

PR #248, opened 2022-11-08
[14.0][IMP] maintenance_request_stage_transition: hook

kluna1998:
Use of _postprocess_view setup modifiers properly and added a hook.


etobella (approved):
Tested locally passing groups and it working like a charm

sanchonuria (approved):

PR #250, opened 2022-11-09
[#15302][14.0][NEW] Add maintenance modules

mmrondon:
- maintenance_checklist
- maintenance_checklist_auto
- maintenance_checklist_plan
- maintenance_checklist_timer


etobella:
Can you make a pr for each module explaining each goal. On a first look, there are some wrong files, readme can be improved and you need to pass precommit.

Thanks!

PR #253, opened 2022-11-23
[15.0][REF] moving function get_relativedelta to method

mymage:
Follow up of #191
Moving get_relativedelta function to MaintenancePlan method


etobella (approved):
/ocabot merge minor

PR #270, opened 2022-12-21
[13.0][IMP] maintenance_timesheet: Add timesheet page from maintenance request form view.

victoralmau:
Add timesheet page from maintenance request form view.

ejemplo

Please @pedrobaeza and @ernestotejeda can you review it?

@Tecnativa TT40881

### DIFF
--- a/base_maintenance/views/maintenance_request_views.xml
+++ b/base_maintenance/views/maintenance_request_views.xml
@@ -6,12 +6,20 @@
         <field name="name">equipment.request.form</field>
         <field name="model">maintenance.request</field>
         <field name="inherit_id" ref="maintenance.hr_equipment_request_view_form" />
+        <field name="priority" eval="999" />
         <field name="arch" type="xml">
             <xpath expr="//sheet/*[1]" position="before">
                 <div class="oe_button_box" name="button_box">
                     <!--Empty, to be inherited by other modules.-->
                 </div>
             </xpath>
+            <field name="description" position="replace">
+                <notebook>
+                    <page name="description_page" string="Description">
+                        <field name="description" />
+                    </page>
+                </notebook>
+            </field>
         </field>
     </record>
 </odoo>
--- a/maintenance_timesheet/__manifest__.py
+++ b/maintenance_timesheet/__manifest__.py
@@ -5,7 +5,7 @@
     "summary": "Adds timesheets to maintenance requests",
     "author": "Odoo Community Association (OCA), Solvos",
     "license": "AGPL-3",
-    "version": "13.0.1.1.0",
+    "version": "13.0.1.2.0",
     "category": "Human Resources",
     "website": "https://github.com/OCA/maintenance",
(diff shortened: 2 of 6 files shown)


pedrobaeza (approved):

base_maintenance/views/maintenance_request_views.xml
```diff
@@ -12,6 +12,13 @@
                     <!--Empty, to be inherited by other modules.-->
                 </div>
             </xpath>
+            <field name="description" position="replace">
```
dalonsod, line 16: With a `replace` in a view, a lower priority for the view is recommended

dalonsod (changes requested):
See comment.

And also,
- Totally agree making `timesheet_total_hours` stored, kanban and tree views will have a better performance. I'm not sure but, a migration hook should be neeeded in this case for older requests and timesheets?
- Timesheets original button box for a request is still present, in fact, I use and like it. Anyway, I don't know if this redundancy is the right approach.
- `base_maintenance` is also changed. A standalone PR or, at least, an individual commit, should be made instead?

base_maintenance/views/maintenance_request_views.xml
victoralmau: Ok, already changed in a separate commit

victoralmau:
>     * Totally agree making `timesheet_total_hours` stored, kanban and tree views will have a better performance. I'm not sure but, a migration hook should be neeeded in this case for older requests and timesheets?
It is not necessary, when updating the module, the compute will define the value.
>     * Timesheets original button box for a request is still present, in fact, I use and like it. Anyway, I don't know if this redundancy is the right approach.
Exactly, I didn't want to remove the smart-button in case someone uses it.
>     * `base_maintenance` is also changed. A standalone PR or, at least, an individual commit, should be made instead?
Ok, I have separated it in another commit.

dalonsod:
> It is not necessary, when updating the module, the compute will define the value.

OK, anyway IMO a good option in this case should be a pre-migration SQL script in order to prevent a heavy amount of maintenance requests. What do you think? cc @pedrobaeza 

Tested in v12, the update query could be as follows:

```
UPDATE maintenance_request
SET timesheet_total_hours=coalesce(tt.total_unit_amount, 0.0)
FROM (
    SELECT mr.id, SUM(aal.unit_amount) total_unit_amount
    FROM maintenance_request mr
    LEFT JOIN account_analytic_line aal ON aal.maintenance_request_id = mr.id
    GROUP BY mr.id
) AS tt
WHERE maintenance_request.id = tt.id
```

pedrobaeza:
I don't think the upgrade time is going to be very huge, but OK for the migration script.

victoralmau:
Migration script added.

victoralmau:
With these changes merged, I will add in `maintenance_timesheet_time_control` the buttons in this tree, or better to do it already here?

dalonsod (approved):
👍

PR #272, opened 2022-12-23
[13.0][FIX] maintenance_plan: Avoid multiple notebooks

victoralmau:
Avoid multiple notebooks (if `maintenance_timesheet` is installed).

**Before**
antes

**After**
despues

Since https://github.com/OCA/maintenance/pull/270

Please @pedrobaeza and @ernestotejeda can you review it?

@Tecnativa TT40881

### DIFF
--- a/maintenance_plan/__manifest__.py
+++ b/maintenance_plan/__manifest__.py
@@ -9,7 +9,7 @@
     "category": "Maintenance",
     "website": "https://github.com/OCA/maintenance",
     "images": [],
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
It's a change of dependencies, but I think it's tolerable.

dalonsod (approved):
Well, for `maintenance_plan` depending on `base_maintenance` is a good point IMO

ernestotejeda (approved):

PR #283, opened 2023-01-13
[13.0][IMP] maintenance_plan: Set planned_hours in request from plan if maintenance_timesheet is installed.

victoralmau:
Set `planned_hours` in request from plan if `maintenance_timesheet` is installed.

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -57,14 +57,15 @@ class MaintenanceEquipment(models.Model):
 
     def _prepare_request_from_plan(self, maintenance_plan, next_maintenance_date):
         team = maintenance_plan.maintenance_team_id.id or self.maintenance_team_id.id
+        request_model = self.env["maintenance.request"]
         if not team:
-            team = self.env["maintenance.request"]._get_default_team_id()
+            team = request_model._get_default_team_id()
 
         description = self.name if self else maintenance_plan.name
         kind = maintenance_plan.maintenance_kind_id.name or _("Unspecified kind")
         name = _("Preventive Maintenance (%s) - %s") % (kind, description)
 
-        return {
+        data = {
             "name": name,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Aren't you mixing `maintenance_plan` and `maintenance_timesheet`?

victoralmau:
> Aren't you mixing `maintenance_plan` and `maintenance_timesheet`?

To prevent a a glue module, I have set the method to be used only if `maintenance_plan` is installed. Do you prefer to do it another way?

pedrobaeza:
That doesn't serve, Víctor. You can't call super over a method that is not defined on its ancestors in the dependency chain. It should be the contrary: in `maintenance_plan`, if `planned_hours` exist, it's filled.

victoralmau:
> That doesn't serve, Víctor. You can't call super over a method that is not defined on its ancestors in the dependency chain. It should be the contrary: in `maintenance_plan`, if `planned_hours` exist, it's filled.

Ok, changes done.

maintenance_plan/models/maintenance_equipment.py
```diff
@@ -80,6 +81,9 @@ def _prepare_request_from_plan(self, maintenance_plan, next_maintenance_date):
             "note": maintenance_plan.note,
             "company_id": maintenance_plan.company_id.id or self.company_id.id,
         }
+        if "planned_hours" in request_model._fields:
```
pedrobaeza, line 85: You always forget to comment why doing these things that will serve in the future. In this case, a comment saying that this field comes from `maintenance_timesheet` and for avoiding a glue module just for that.

victoralmau:
Ping @pedrobaeza

pedrobaeza (approved):

CarlosRoca13 (approved):
LGTM

PR #287, opened 2023-01-29
[14.0][IMP] maintenance_checklist_timer

mmrondon:


maintenance_checklist_timer/models/maintenance_checklist.py
```diff
@@ -0,0 +1,52 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import _, models
+
+
+class MaintenanceChecklist(models.Model):
+    _name = "maintenance.checklist"
+    _inherit = ["maintenance.checklist", "timer.mixin"]
```
mamcode, line 7: This Mixin belongs to a enterprise module :(
https://github.com/odoo/enterprise/blob/14.0/timer/models/timer_mixin.py

mamcode (changes requested):
Thanks for this module it is useful but do you think there is an alternative to using the Mixin timer.mixin ?

PR #307, opened 2023-02-27
[13.0][IMP] maintenance_plan: Skip assigned mail + Activity mail of the requests created from plans

victoralmau:
Skip assigned mail + Activity mail of the requests created from plans

Please  @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT41870

### DIFF
--- a/maintenance_plan/i18n/es.po
+++ b/maintenance_plan/i18n/es.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 12.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-02-20 07:51+0000\n"
-"PO-Revision-Date: 2023-02-20 08:51+0100\n"
+"POT-Creation-Date: 2023-03-01 13:30+0000\n"
+"PO-Revision-Date: 2023-03-01 14:31+0100\n"
 "Last-Translator: Víctor Martínez <victor.martinez@tecnativa.com>\n"
 "Language-Team: none\n"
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

pedrobaeza:
@dalonsod what do you think about this change for avoid hammering people with a lot of assignation mails?

CarlosRoca13 (approved):

dalonsod:
This improvement looks good... is it possible to be this new behavior parametrized for each plan (and set as "do not notify to follower when creating requests" as default)? Then, current functionality could be still achieved if needed

victoralmau:
Changes done (new field created in plan to be able to configure it).

dalonsod:
Not blocking,

- New parameter is set to `False` as default. Maybe it could be set to `True` (suggested behavior, the original one for this improvement you looked for), so be repeatedly notified should be strictly chosen.
- `po` files, should be excluded from commit?

victoralmau:
> Not blocking,
> 
>     * New parameter is set to `False` as default. Maybe it could be set to `True` (suggested behavior, the original one for this improvement you looked for), so be repeatedly notified should be strictly chosen.
Changed to True by default.
>     * `po` files, should be excluded from commit?
I think it is correct, the translation is already added directly without the need to do it in weblate.

pedrobaeza:
About changing PO files, there may be the possibility of a conflict because meanwhile, there's a change in Weblate, or not all changes are pushed back from Weblate to GitHub, but as the translation coverage is good, the chances are very low, and this pull request life is short.

PR #309, opened 2023-03-01
[14.0][FIX] maintenance_account /maintenance_purchase: check before creation

etobella:
This fixes a bug when you create equipments from a purchase/invoice with two lines, with products with the same category but it was not used before. Two identical categories would have been created.

### DIFF
--- a/maintenance_account/models/account_move.py
+++ b/maintenance_account/models/account_move.py
@@ -100,11 +100,14 @@ class AccountMoveLine(models.Model):
 
     def _set_equipment_category(self):
         if not self.equipment_category_id:
-            # Prevent error if user does not have permission to create equipments
             category_model = self.env["maintenance.equipment.category"].sudo()
-            category = category_model.with_company(self.company_id).create(
-                self._prepare_equipment_category_vals()
+            category = fields.first(
+                self.product_id.product_tmpl_id.categ_id.equipment_category_ids
             )
+            if not category:
+                category = category_model.create(
+                    self._prepare_equipment_category_vals()
+                )
             self.equipment_category_id = category.id
 
     def _prepare_equipment_vals(self):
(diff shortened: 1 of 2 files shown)


maintenance_account/models/account_move.py
```diff
@@ -98,8 +98,15 @@ def _prepare_equipment_category_vals(self):
...
-            category = category_model.create(self._prepare_equipment_category_vals())
+            category = fields.first(
+                self.product_id.product_tmpl_id.categ_id.equipment_category_ids
+            )
+            if not category:
+                category = category_model.create(
+                    self._prepare_equipment_category_vals()
+                )
             self.equipment_category_id = category.id
+            category = category_model.create(self._prepare_equipment_category_vals())
```
victoralmau, line 109:
```suggestion
```

victoralmau (changes requested):
Thank you, can you improve the tests to check that this does not happen in the future?

victoralmau (changes requested):
Reviewing this in detail I see that the problem is the depends, this would be the change in v15: https://github.com/OCA/maintenance/pull/312/commits/bceba8dca1407bbbe89d8505c57f4ca68e7e299d

In `maintenance_purchase` the change would be similar.

pedrobaeza:
Please don't do that. If you put such depends, and you change any equipment in the product category, it will recompute all the past moves.

victoralmau:
> Please don't do that. If you put such depends, and you change any equipment in the product category, it will recompute all the past moves.

Ok, I have changed the approach in https://github.com/OCA/maintenance/pull/312

maintenance_account/models/account_move.py
pedrobaeza: This one is duplicated, isn't it?

pedrobaeza:
@etobella please resolve conflicts and remove the duplicated creation for proceeding with the merge.

pedrobaeza (approved):
/ocabot merge patch

PR #310, opened 2023-03-01
[13.0][FIX] maintenance_account: Avoid error when post the invoice if the user does not have permissions to create maintenance records.

victoralmau:
Avoid error when post the invoice if the user does not have permissions to create maintenance records.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT41917

### DIFF
--- a/maintenance_account/models/account_move.py
+++ b/maintenance_account/models/account_move.py
@@ -19,7 +19,8 @@ class AccountMove(models.Model):
 
     def action_post(self):
         super().action_post()
-        equipment_model = self.env["maintenance.equipment"]
+        # Prevent error if user does not have permission to create equipments
+        equipment_model = self.env["maintenance.equipment"].sudo()
         for move in self.filtered(lambda r: r.is_purchase_document()):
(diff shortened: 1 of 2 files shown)


maintenance_account/models/account_move.py
```diff
@@ -19,7 +19,8 @@ def unlink(self):
 
     def action_post(self):
         super().action_post()
-        equipment_model = self.env["maintenance.equipment"]
+        # Prevent error if user does not have permission to create equipments
+        equipment_model = self.env["maintenance.equipment"].sudo()
```
pedrobaeza, line 23: Pass also company of the invoice for multi-company environments, with context `force_company`. You need to do it per document.

pedrobaeza:
Please don't do 2 commits. The problem of the multi-company arrives when you do `sudo`, so it's related. In a non-sudo environment, everything goes to the proper company. And you should do it in both places, not in only one.

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #333, opened 2023-04-21
[14.0][FIX] maintenance_account: Add the field equipment_ids

etobella:
We want to add it in order to avoid that the information is filled when using the purchase field.


sanchonuria (approved):

kluna1998 (approved):

victoralmau:
Sorry, can you explain the use case for the need to add that field?

etobella:
If the field is not there, the invoices will be related to no purchases once we create the invoice from the purchase. So, the equipment will be generated from the purchase and again from the invoice

victoralmau (approved):

PR #336, opened 2023-04-26
[16.0][MIG] maintenance_request_employee

mmrondon:


dalonsod (commented):
Functional review 👍 

Only two improvement suggestions (if made, in a separated commit, so can be backported to v15):
- enable `employee_ids` field for grouping in `search` view.
- add that field in `tree` view (with `many2many_tags`)

PR #342, opened 2023-06-21
[16.0][MIG] maintenance_product: Migration to 16.0

kaynnan:
cc @marcelsavegnago 

The changes in `_onchange_product_id` were made because the error message indicated that the `product.supplierinfo` object didn't have a `name` attribute. Therefore, the code was modified to assign the value to `partner_id`, and it had the same effect.


marcelsavegnago (approved):

maintenance_product/views/maintenance_equipment_view.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="pm_maintenance_equipment_form_view" model="ir.ui.view">
+        <field name="name">pm.maintenance.equipment.form.view</field>
+        <field name="model">maintenance.equipment</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_view_form" />
+        <field name="arch" type="xml">
+            <field name="category_id" position="after">
+                <field name="product_category_id" invisible="1" />
+                <field name="product_id" />
```
christian-ramos-tecnativa:
```suggestion
                <field name="product_id" context="{'default_maintenance_ok': True}" />
```
It would be nice to add context in the form field, so if you create a product on the fly it will meet the domain requirements. Or just disable the create option.
For the rest LGTM

kaynnan: Thank you for the review. I followed your suggestion and implemented that feature.

christian-ramos-tecnativa (approved):

etobella:
Can you join your two commits?

/ocabot migration maintenance_product

kaynnan:
> Can you join your two commits?
> 
> 
> 
> /ocabot migration maintenance_product

Thanks for the suggestion, i made the squash commits

etobella:
There is still two commits (maybe there were three and I didn't realize :disappointed: )

kaynnan:
> There is still two commits (maybe there were three and I didn't realize :disappointed: )

Sorry, typo here and the squash command didn't work, i now believe the history commit is correct, only pre-commit and migration

etobella (approved):

PR #350, opened 2023-07-11
[15.0][ADD] maintenance_security: New module

victoralmau:
New module

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT43038


maintenance_security/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module displays the Maintenance menu only to users with maintenance permissions and limits the users to whom the different maintenance fields (equipment, requests, team members and category managers) can be assigned.
```
pedrobaeza: Remember to split the lines into 88 cols the same as code:

```suggestion
This module limits the Maintenance menu to users with any of the maintenance groups.

It also removes the assignation mail when a user with no permission is assigned to an equipment, as if not, the mail will contain a link to it.
```

pedrobaeza (changes requested):
Instead of restricting which users can be assigned, the module should restrict the automatic assignation mail when someone which is not from that security groups is assigned.

victoralmau:
Changes done (different approach).

CarlosRoca13 (approved):

maintenance_security/views/maintenance_views.xml
pedrobaeza, line 1: Do you need to set all of these menus? Why not only the root one?

maintenance_security/models/mail_followers.py
```diff
@@ -0,0 +1,37 @@
...
+    def _add_followers(
+        self,
+        res_model,
+        res_ids,
+        partner_ids,
+        subtypes,
+        check_existing=False,
+        existing_policy="skip",
+    ):
+        """We need to prevent a partner without Maintenance permission from being
```
pedrobaeza, line 18: This is preventing to be added as follower, but not the special case that is embedded in `mail` module, that when you set the field `user_id`, a mail with subject "You have been assigned to" is sent. Please restrict also this one.

victoralmau: Sorry, probably have to explain more the reason for this. `maintenance.equipment` does not have the `user_id` field, similar field is https://github.com/odoo/odoo/blob/15.0/addons/maintenance/models/maintenance.py#L117 and that field has `tracking=true`. 

With each change in that field https://github.com/odoo/odoo/blob/15.0/addons/maintenance/models/maintenance.py#L212 becomes a follower of a specific subtype https://github.com/odoo/odoo/blob/15.0/addons/maintenance/models/maintenance.py#L90

The email that is sent is for the `mail.tracking` record of this change, and for this reason the solution has had to be this, to prevent it from being added as a follower of that subtype.

pedrobaeza (approved):
/ocabot merge nobump

PR #128, opened 2021-02-12
[13.0][MIG] maintenance_equipment_custom_info

dalonsod:
Standard migration

PR #89, opened 2020-06-01
[12.0][ADD] maintenance_equipment_custom_info

dalonsod:
This module extends maintenance equipments to allow having custom properties on them

PR #356, opened 2023-09-06
[ADD] maintenance_equipment_custom_info

etobella:
This modules integrate base_custom_info with maintenance equipments


dalonsod:
Hello @etobella related to this new addon, have you checked OCA/server-tools#2330 ?

etobella:
Nope, I am not sure why this could happen :thinking:

etobella:
@dalonsod May I ask for your review :wink:

dalonsod:
Code review, noticed that no security group is defined, like Contacts one:

https://github.com/OCA/server-tools/blob/eac8defb5cded460ce8e13e250a8d7855d937d92/base_custom_info/security/res_groups_security.xml#L9-L15

Would you consider creating it?

etobella:
The contacts group is created in order to know if the implementation has sense for you on contacts (as it comes with the base module). This one doesn't need it as you assume that you already want it. on equipments.

dalonsod:
OK, I know... other possibility should be having that group in order to allow some users maintaining custom info for equipments, and disabling it for others. But it's an optional improvement that could be added in a later PR anyway.

And, should be added `oca-dependencies.txt` reference to `server-tools` repo, or is it unnecessary?

dalonsod (approved):

etobella:
With the change if pip install, there is no need of oca_dependencies 😄

ChristianSantamaria (approved):
LGTM 👍

dalonsod:
@etobella this is embarrassing... I helped you with this PR, and I've noticed that it was already done in `12.0` and `13.0`... by me #89 #128 😅

Looking over `14.0` implementation seems to be coherent with previous implementation, so migration problems should not raise. But, related to addon authorship, should I be added? I'm not worried about it anyway...

etobella:
No!!! I  didn't know!!! Sorry.... Options: I can migrate it on 15 and restore your history. Would that work for you?

For 14, I would recommend to add you as author, because it is obviously your work. Can you make the PR?

dalonsod:
> No!!! I didn't know!!! Sorry.... Options: I can migrate it on 15 and restore your history. Would that work for you?

:+1: 

> For 14, I would recommend to add you as author, because it is obviously your work. Can you make the PR?

Done, #382

PR #357, opened 2023-09-06
[14.0][IMP] maintenance_plan: Create requests from a domain

etobella:
This PR Allows us to create several requests from a single plan using a domain to set several equipments.

For example, I create a plan for callibration of all the equipments for category Monitor. Then, at the specified time, for all the active monitors in my sistem, a callibration action will be generated.

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -20,6 +20,10 @@ class MaintenanceEquipment(models.Model):
         string="Maintenance Plan Count",
         store=True,
     )
+    search_maintenance_plan_count = fields.Integer(
+        compute="_compute_search_maintenance_plan_count",
+        string="Maintenance All Plan Count",
+    )
     maintenance_team_required = fields.Boolean(compute="_compute_team_required")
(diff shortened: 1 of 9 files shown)


LoisRForgeFlow:
@etobella I think this can be included in the base maintenance_plan module, I don't see any problem for existing instalation, you can just not use the feature.

etobella:
Tests are :green_apple: @LoisRForgeFlow May I ask for your review?

LoisRForgeFlow (commented):
It is a really interesting improvement!

One suggestion to improve UX. What do you think about adding a computed field that links domain based plans with applicable equipments in a way that you can find from an equipment which plan would apply to it?

etobella:
It would be hard to manage it by a widget (problems with creation), but, I could do a button that opens all the plans of an equipment. WDYT?

LoisRForgeFlow:
> It would be hard to manage it by a widget (problems with creation), but, I could do a button that opens all the plans of an equipment. WDYT?

I think that would be great!

LoisRForgeFlow (approved):
LGTM!

PR #358, opened 2023-09-08
[14.0][ADD] maintenance_location

etobella:
This module allows to manage a hierarchy of locations for Maintenance

We are not reusing stock.locations as they have different levels


kluna1998 (approved):
LGTM!

PR #360, opened 2023-09-23
[14.0][FIX] maintenance_plan: Use language on cron creation

etobella:
Without this change, we might have problems when creating the request as data might differ from what you write (for example, if we come from a copied record).

Issued is raised because cron are executed with no lang.

The problems comes from using a field that is translatable

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -61,7 +61,11 @@ class MaintenanceEquipment(models.Model):
         if not team_id:
             team_id = request_model._get_default_team_id()
 
-        description = self.name if self else maintenance_plan.name
+        description = (
+            self.with_context(lang=self.env.user.lang).name
+            if self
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It's weird to use sometimes the maintenance plan name and other times the equipment name.

etobella:
I understand your concerns, but it is already being used, I am not sure if we should change the behavior directly...

pedrobaeza:
Yes, I think it can be improved. Maybe a compose name between equipment name and a "suffix"?

etobella:
But the maintenance request final name will be as you ask:
`_("Preventive Maintenance (%s) - %s") % (kind, description)`
where kind is used to split the kind of preventive request we are setting.  And description will be the name.

pedrobaeza:
OK, as you prefer

mamcode (changes requested):
Looks good, thank you.

mamcode (approved):

PR #362, opened 2023-10-02
[ADD] maintennace_category_hierarchy

etobella:
This modules allows to manage categories as a parent child relationship

@eantones


eantones:
LGTM!

PR #365, opened 2023-10-19
[14.0][FIX] maintenance_purchase: When we copy a purchase, the equipemnts shouldn't be maintained

etobella:


pedrobaeza (approved):
Indeed, please fw-port it

/ocabot merge patch

etobella:
Well, right now it is  only available on 14.0, so all is done :smile:

pedrobaeza:
> Well, right now it is only available on 14.0, so all is done 😄

It's my wildcard answer when merging, for not forgetting the patch in lower versions. Thanks for it!

PR #371, opened 2023-10-31
[15.0][IMP] maintenance_plan: Add freeze_time to prevent errors in tests

victoralmau:
Add `freeze_time` to prevent errors in tests.

@Tecnativa

### DIFF
--- a/maintenance_plan/tests/test_maintenance_plan.py
+++ b/maintenance_plan/tests/test_maintenance_plan.py
@@ -1,8 +1,6 @@
 # Copyright 2017 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from datetime import timedelta
-
 from dateutil.relativedelta import relativedelta
 
 from odoo import _, fields
@@ -11,6 +9,11 @@ from .common import TestMaintenancePlanBase
 
 
 class TestMaintenancePlan(TestMaintenancePlanBase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.today_date = fields.Date.from_string("2023-01-25")
+
     def test_name_get(self):
         self.assertEqual(
             self.maintenance_plan_1.name_get()[0][1],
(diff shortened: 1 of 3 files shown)


maintenance_plan/tests/test_maintenance_plan.py
```diff
@@ -145,7 +145,9 @@ def test_generate_requests(self):
             self.assertEqual(
                 fields.Date.to_date(req.schedule_date), request_date_schedule
             )
-            request_date_schedule = request_date_schedule + relativedelta(months=1)
+            request_date_schedule = request_date_schedule + relativedelta(
```
pedrobaeza: Why putting relative dates instead of absolute ones?

pedrobaeza (commented):
I don't think `month` interval should be `last_day_of_month` option. Anyway, if you put the start date on the 31st, relativedelta + 1 month is the last day, isn't it?

maintenance_plan/tests/test_maintenance_plan.py
victoralmau: Maintenance plans are currently created using today's date.

Another possibility would be to change all tests to run on specific dates and check specific dates.

pedrobaeza: You can use freezegun for having fixed dates.

maintenance_plan/tests/test_maintenance_plan.py
```diff
@@ -1,16 +1,21 @@
...
 
 from .common import TestMaintenancePlanBase
 
 
+@freeze_time("2023-01-31")
 class TestMaintenancePlan(TestMaintenancePlanBase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.today_date = fields.Date.from_string(fields.Date.today())
```
pedrobaeza: Hardcode dates instead of using this and having to check up which day it is:

```suggestion
        cls.today_date = fields.Date.from_string("2023-01-31")
```

and maybe storing this date doesn't have too much sense, but put it directly in asserts.

pedrobaeza (commented):
I still feel that changing the behavior of current interval option is not OK. You should add a new one "End of month".

victoralmau:
> I still feel that changing the behavior of current interval option is not OK. You should add a new one "End of month".

You are right, the previous change (end of month) should be an option to be defined to avoid the data not being as expected.

Finally I have left only the `freeze_time` change to make the tests more resilient.

pedrobaeza:
Please check this also: https://github.com/orgs/OCA/discussions/122

pedrobaeza (approved):
/ocabot merge patch

PR #376, opened 2023-12-11
[FIX] maintenance_equipment_scrap: Scrap date is hidden

etobella:


mamcode (approved):
Looks good and I agree with the change. 

I only have the doubt why Odoo makes the scrapping date only visible in developer mode by default...

etobella:
It is something they never changed from the origin

https://github.com/odoo/odoo/commit/f63d2d5135c0a6e3d032082ec8bfce16c40add69

dalonsod (approved):
:+1:

PR #377, opened 2023-12-13
[FIX] maintenance_request_purchase: Do not copy fields

etobella:
When copying a record, we should not store this data, at least it has no sense IMO :smile:


kluna1998 (approved):
LGTM!

PR #378, opened 2023-12-13
[ADD] maintenance_equipment_meter

etobella:
This module allows us to integrate a meter (Odometer, printed pages counter...) on equipments. Data can be fulfilled in equipments directly or in requests.

Approach is similar to fleet odometer, but for equipments


maintenance_equipment_meter/views/maintenance_equipment_meter.xml
```diff
@@ -0,0 +1,86 @@
...
+        <field name="domain">[]</field>
+        <field name="context">{}</field>
+    </record>
+
+    <record model="ir.ui.menu" id="maintenance_equipment_meter_menu">
+        <field name="name">Maintenance Equipment Meter</field>
+        <field name="parent_id" ref="xmlid of parent menu"/>
+        <field name="action" ref="maintenance_equipment_meter_act_window"/>
+        <field name="sequence" eval="16"/>
+    </record>
```
mamcode: It seems that the intention of this commented code was to have a main menu to access all the meter records of all the equipments from a main menu. It would be nice if we can have this. 
By the way, in this action I would group the records by equipment.

mamcode (commented):
Looks good to me. Thank you for this new module.

mamcode (approved):

PR #380, opened 2024-01-17
[FIX] maintenance_account action_post missing return

sergiocorato:
Causes problems with modules like account_financial_risk that return a window action.


victoralmau (approved):

mamcode (approved):
LGTM thanks

PR #382, opened 2024-02-20
[14.0][UPD] maintenance_equipment_custom_info: restore authorship

dalonsod:
@etobella please check if it's ok


etobella (approved):
Sorry for the mistake

Thanks for fixing it

PR #386, opened 2024-03-04
[15.0][IMP] maintenance_account: Create equipment description from move line name

carolinafernandez-tecnativa:
- The first line of the name from account.move should go to the name of the equipment.
- The rest of the lines go to the equipment description

@Tecnativa
TT47802
@pedrobaeza @victoralmau

PR #387, opened 2024-03-05
[16.0][IMP] maintenance_account: Create equipment description from move line name

carolinafernandez-tecnativa:
- Forward-port of https://github.com/OCA/maintenance/pull/386
- The first line of the name from account.move should go to the name of the equipment.
- The rest of the lines go to the equipment description

@Tecnativa
TT47802

@pedrobaeza @victoralmau


pedrobaeza:
Indicate that this is a forward-port of ...

Why is it draft?

carolinafernandez-tecnativa:
> Indicate that this is a forward-port of ...
> 
> Why is it draft?

I change state when everything is fine. Now its ready for review :)
Indicated foward-port

maintenance_account/models/account_move.py
```diff
@@ -113,13 +114,20 @@ def _set_equipment_category(self):
...
             "move_line_id": self.id,
-            "name": self.product_id.name,
+            "name": equipment_name,
             "product_id": self.product_id.id,
             "category_id": self.equipment_category_id.id,
             "assign_date": self.move_id.date,
             "effective_date": self.move_id.date,
             "partner_id": self.move_id.partner_id.id,
             "partner_ref": self.move_id.ref,
+            "note": description,
```
pedrobaeza, line 133: Shouldn't this be converted to HTML?

maintenance_account/tests/test_maintenance_account.py
```diff
@@ -60,6 +62,7 @@ def _create_invoice(self, move_type="in_invoice"):
         move_form.invoice_date = fields.Date.from_string("2000-01-01")
         with move_form.invoice_line_ids.new() as line_form:
             line_form.product_id = self.product_a
+            line_form.name = "Product A \nTest description product A"
```
pedrobaeza: Beware with the extra space:

```suggestion
            line_form.name = "Product A\nTest description product A"
```

maintenance_account/tests/test_maintenance_account.py
```diff
@@ -92,7 +95,9 @@ def test_invoice_action_post_equipment_1(self):
         self.assertEqual(len(line_a.equipment_ids), 2)
         self.assertEqual(len(line_b.equipment_ids), 0)
         equipment = fields.first(equipments)
-        self.assertEqual(equipment.name, self.product_a.name)
+        name_list = line_a.name.split("\n")
+        self.assertEqual(equipment.name, name_list[0])
+        self.assertEqual(html2plaintext(equipment.note), name_list[1])
```
pedrobaeza: Don't test code. Test fixed data:

```suggestion
        self.assertEqual(equipment.name, "Product A")
        self.assertEqual(equipment.note, "<p>Test description product A</p>")
```

pedrobaeza (changes requested):
And I have now the doubt if you need to do the same of converting to HTML in v15

maintenance_account/models/account_move.py
carolinafernandez-tecnativa: It is not necessary as self.name is only text for this case.

pedrobaeza: Yes, but `description` is HTML, so passing `Line 2\nLine 3` is incorrect. You should convert it to `Line 2<br/>Line 3` for example.

carolinafernandez-tecnativa: done

maintenance_account/models/account_move.py
```diff
@@ -113,13 +115,20 @@ def _set_equipment_category(self):
             self.equipment_category_id = category.id
 
     def _prepare_equipment_vals(self):
+        equipment_name = self.name
+        description = False
+        if "\n" in self.name:
+            lf_index = self.name.index("\n")
+            equipment_name = self.name[:lf_index]
+            description = html_escape(self.name[lf_index + 1 :])
```
pedrobaeza: I would say `html_escape` is not for converting text to HTML.

pedrobaeza: https://github.com/odoo/odoo/blob/ed890131f5bff8264e152c77ae5d3757f2dd7bfb/odoo/tools/misc.py#L1332

https://markupsafe.palletsprojects.com/en/2.1.x/

carolinafernandez-tecnativa: could you please check it now?

pedrobaeza: It's still not valid:

imagen

But I wonder if you have put a test and it's green, is it because v16 makes the conversion automatically, or is it because the test is not being executed?

In v15, I have just checked, and it's incorrect. Look at these 2 screenshots that are self-explanatory:

imagen

imagen

carolinafernandez-tecnativa: Could you please check it now?
I do not know why there where some error test and was checked as green over here...
On v16 it is transform automatically, but if you add \n on text wont recognise, thats why i added a replace on field.
Im checking why user_type_id are not found...

maintenance_account/models/account_move.py
```diff
@@ -113,13 +114,20 @@ def _set_equipment_category(self):
             self.equipment_category_id = category.id
 
     def _prepare_equipment_vals(self):
+        equipment_name = self.name
+        description = False
+        if "\n" in self.name:
+            lf_index = self.name.index("\n")
+            equipment_name = self.name[:lf_index]
+            description = self.name[lf_index + 1 :].replace("\n", "<br>")
```
pedrobaeza: Please use `plaintext2html` instead.

maintenance_account/tests/test_maintenance_account.py
```diff
@@ -60,6 +66,7 @@ def _create_invoice(self, move_type="in_invoice"):
         move_form.invoice_date = fields.Date.from_string("2000-01-01")
         with move_form.invoice_line_ids.new() as line_form:
             line_form.product_id = self.product_a
+            line_form.name = "Product A\nTest description product A\n<b>B</b>"
```
pedrobaeza: You can remove `<b>` from here, as it can be misleading of the real purpose.

maintenance_account/tests/test_maintenance_account.py
```diff
@@ -1,17 +1,23 @@
 # Copyright 2022-2024 Tecnativa - Víctor Martínez
+# Copyright 2024 Tecnativa - Carolina Fernandez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
 from odoo import fields
 from odoo.tests import Form, common, new_test_user
 from odoo.tests.common import users
 
-from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
```
pedrobaeza: Don't change this. It was OK.

pedrobaeza (approved):
/ocabot merge minor

PR #390, opened 2024-03-11
16.0 [IMP] maintenance_timesheet: use create_multi in account.analytic.line

IT-Ideas:


maintenance_timesheet/models/hr_timesheet.py
```diff
@@ -16,31 +16,38 @@ def onchange_maintenance_request_id(self):
...
     def unlink(self):
-        for timesheet in self.filtered(lambda x: x.maintenance_request_id):
-            self._check_request_done(timesheet.maintenance_request_id.id)
+        self._check_request_done(
+            self.filtered(lambda x: x.maintenance_request_id).maintenance_request_id.ids
+        )
         return super().unlink()
 
-    def _check_request_done(self, request_id):
+    def _check_request_done(self, request_ids):
```
pedrobaeza: You shouldn't change the API of this method, as there can be other modules overriding it. Instead, just call N times the method.

IT-Ideas: You are right. On the other hand this will trigger a query for each record. We should change this in the future.

IT-Ideas: I'll propose an alternative that satisfies both worlds... I'll let me know what you think

IT-Ideas:
@pedrobaeza what do you think?

maintenance_timesheet/models/hr_timesheet.py
```diff
@@ -16,11 +16,15 @@ def onchange_maintenance_request_id(self):
...
-        return super().create(values)
+    @api.model_create_multi
+    def create(self, vals_list):
+        maintenance_request_ids = [
+            vals.get("maintenance_request_id")
+            for vals in vals_list
+            if vals.get("maintenance_request_id")
+        ]
+        self._check_requests_done(maintenance_request_ids)
+        return super().create(vals_list)
```
pedrobaeza: Please don't complicate API with 2 methods with just an extra `s` (very easy to be typo error), and don't complicate logic. It's easier this way:

```suggestion
        for vals in vals_list:
            if vals.get("maintenance_request_id"):
                self._check_requests_done(vals["maintenance_request_id"])
        return super().create(vals_list)
```

IT-Ideas: Hello @pedrobaeza , the problem to me is really that [`https://github.com/OCA/maintenance/pull/390/files#diff-ce85fd70c56eb1115afe99031137572f69d8b227218faad2d9119f9d5fcde50dR55`](https://github.com/OCA/maintenance/pull/390/files#diff-ce85fd70c56eb1115afe99031137572f69d8b227218faad2d9119f9d5fcde50dL43) will make a distinct select for each record... The create_multi has been made in order to prevent those situations so I personally find that 'just' making a loop around the previous implementation is a bit sad... 

I would rather change the name in order to avoid typo issue and keep my implementation. What name would you propose?

lmignon: @IT-Ideas We could keep the same method and document that it could be called with a single id or with a list of ids...

maintenance_timesheet/models/hr_timesheet.py
```diff
@@ -32,15 +36,24 @@ def write(self, values):
...
-            self._check_request_done(timesheet.maintenance_request_id.id)
+        self._check_requests_done(
+            self.filtered(lambda x: x.maintenance_request_id).maintenance_request_id.ids
+        )
         return super().unlink()
 
     def _check_request_done(self, request_id):
         """
         Editing a timesheet related to a finished request is forbidden.
         """
```
lmignon:
```suggestion
    def _check_request_done(self, request_id: int | list[int]):
        """
        Editing a timesheet related to a finished request is forbidden.
        """
        request_ids = [request_id] if isinstance(request_id, int) else request_id
```

maintenance_timesheet/models/hr_timesheet.py
```diff
@@ -32,15 +36,24 @@ def write(self, values):
...
-        if self.env["maintenance.request"].browse(request_id).stage_id.done:
+        self._check_requests_done([request_id])
+
+    def _check_requests_done(self, request_ids):
+        """
+        Editing a timesheet related to a finished request is forbidden.
+        """
```
lmignon:
```suggestion

```

lmignon (changes requested):

pedrobaeza (approved):
OK, the polymorphic option is not my best option (due to the argument variable name), but I don't block this more.

lmignon (approved):
As of the polymorphic approach is documented, it's the bet solution to take benefit of perf improvement without introducing  duplicate code. The other option would be to keep only one method taking a list of ids as arg and update the existing call to the method accordingly...

PR #395, opened 2024-04-10
[16.0][IMP] maintenance_plan: remove python libs included in odoo requirements

CRogos:
same story as https://github.com/OCA/reporting-engine/pull/865

https://github.com/odoo/odoo/blob/16.0/requirements.txt


etobella (approved):
@CRogos this is ready for review? It seems correct for  me

CRogos:
Thank you for asking. I think we leave it as is. There is a huge discussion for both ways and there is no optimal solution yet.
https://github.com/OCA/queue/pull/530

PR #414, opened 2024-07-23
[17.0][MIG] maintenance_plan: Migration to 17.0

FernandoRomera:


FernandoRomera:
@pedrobaeza @astirpe @etobella 
In version 17 they have removed the cron to generate maintenance requests and replaced it with activities. Does anyone have any suggestions for adapting the code.

pedrobaeza:
/ocabot migration maintenance_plan

I think we should add a cron ourselves to continue creating the corresponding requests.

etobella:
I think we should restore the cron. The idea they are providing is not bad and improves their last version, but I think it is not enough.

FernandoRomera:
@pedrobaeza 
Add cron.

AntoniRomera (approved):
LGTM

MRomeera (approved):
LGTM

FernandoRomera:
@pedrobaeza @astirpe @etobella
Can you merge this PR, please?

pedrobaeza:
A PSC or maintainer should confirm, like @etobella

maintenance_plan/__manifest__.py
```diff
@@ -3,23 +3,24 @@
...
     "summary": "Extends preventive maintenance planning",
-    "version": "16.0.1.0.0",
+    "version": "17.0.1.0.0",
     "author": "Camptocamp SA, ForgeFlow, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Maintenance",
     "website": "https://github.com/OCA/maintenance",
     "images": [],
-    "depends": ["base_maintenance"],
+    "depends": ["base_maintenance", "maintenance"],
```
etobella: I think maintenance is not required (base_maintenance depends on it)

https://github.com/OCA/maintenance/blob/17.0/base_maintenance/__manifest__.py#L12

maintenance_plan/hooks.py
```diff
@@ -23,17 +21,17 @@ def post_init_hook(cr, registry):
 
         for equipment in equipments:
             request = equipment.maintenance_ids.filtered(
-                lambda r: r.maintenance_type == "preventive"
+                lambda r, equipment=equipment: r.maintenance_type == "preventive"
```
etobella, line 24: Why this change?

FernandoRomera: B023 Function definition does not bind loop variable `equipment

<https://docs.astral.sh/ruff/rules/function-uses-loop-variable/>`

maintenance_plan/__manifest__.py
FernandoRomera: Sorry, my mistake.

FernandoRomera: @etobella 
Done.

maintenance_plan/hooks.py
etobella: I see, it is a new requirement from ruff :+1:

etobella (approved):

victoralmau:
Analyzing the `maintenance` module in v17 to add the migration script to OpenUpgrade I see that what is indicated here https://github.com/OCA/maintenance/pull/414#issuecomment-2247330930 is not exactly correct, because there is a recurrence field (`recurring_maintenance` and extra fields). 

When creating now a request of preventive type and set recurrence (7 days for example), when the status is changed to one defined as “_completed_” (`done` field) a new request will be created (copy) with the corresponding date (`schedule_date`) https://github.com/odoo/odoo/blob/17.0/addons/maintenance/models/maintenance.py#L337 and in the new one we can maintain that recurrence, change it, etc (the corresponding activity is also added). 

Having clarified the above I am not sure that this module is necessary. I comment this because the approach of the migration script will be different (define the corresponding data to use the odoo recurrence or use this module).

victoralmau:
What do you think about my previous comment (OpenUpgrade scripts will depend on it because maybe this module is not necessary)? Ping @carlos-lopez-tecnativa

maintenance_plan/views/maintenance_plan_views.xml
```diff
@@ -0,0 +1,204 @@
...
+<odoo>
+    <record id="hr_equipment_request_action_from_plan" model="ir.actions.act_window">
+        <field name="name">Maintenance Requests</field>
+        <field name="res_model">maintenance.request</field>
+        <field name="binding_model_id" ref="model_maintenance_plan" />
+        <field name="view_mode">kanban,tree,form,pivot,graph,calendar</field>
+        <field name="context">{
+            'default_maintenance_plan_id': id,
+        }</field>
+        <field name="domain">[('maintenance_plan_id', '=', id)]</field>
```
carlos-lopez-tecnativa, line 11: image

maintenance_plan/models/maintenance_plan.py
```diff
@@ -0,0 +1,260 @@
...
+                    plan.id,
+                    plan.name
+                    or _(
+                        "Unnamed %(kind)s plan (%(eqpmt)s)",
+                        kind=plan.maintenance_kind_id.name or "",
+                        eqpmt=plan.equipment_id.name,
+                    ),
+                )
+            )
+        return result
```
carlos-lopez-tecnativa, line 135: Not blocking, but this can be replaced by `_compute_display_name` according to the migration guidelines.
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

carlos-lopez-tecnativa (commented):
> What do you think about my previous comment (OpenUpgrade scripts will depend on it because maybe this module is not necessary)? Ping @carlos-lopez-tecnativa

Thanks for the ping, @victoralmau. I tested this module and have some observations related to its migration.

Regarding recurrency, I believe this module provides a functionality that Odoo does not currently have—the ability to create maintenance requests in advance based on the `Planning Horizon`. In Odoo, you can only create the next maintenance request when the current one moves to a stage marked as "Done." However, for preventive maintenance in some contexts, it's necessary to generate all future maintenance requests in advance.

I think this module is still valuable but could benefit from some enhancements. For instance, it would be helpful if all maintenance requests related to the recurrency were visible within the maintenance.request form view, similar to how related tasks are displayed in project.task.

What do you think about this approach?

image

carlos-lopez-tecnativa:
Please let me know the progress on this module's evolution. I plan to start the migration of the `maintenance_plan_employee` module, which depends on this one. @pedrobaeza

pedrobaeza:
@carlos-lopez-tecnativa do the PR improving this module for not getting blocked.

victoralmau:
My conclusions in summary after discussing it with my colleague @carlos-lopez-tecnativa 

**1- Keep the OCA module**

Pros: 
- Previous users still work with the same approach.
- All stock requests for a plan are already created.
- No change in Openupgrade is required (https://github.com/OCA/OpenUpgrade/pull/4654).

Cons:
- We are not taking advantage of the approach (fields) of recurrence in requests, causing confusion for whoever uses them and/or when they discover it.
- We are adding code (and the module itself) `maintenance_plan` that core has already contemplated otherwise.
- It is necessary to make change in Openupgrade https://github.com/OCA/OpenUpgrade/pull/4654

**2- Do NOT use the OCA module**

Pros:
- No inconvenience for users who did not know it until now.
- You can create recurrences (plans) from the first request.
- Not all future requests are created at once, the next one will be created when the previous one is finished (more flexibility in the change of recurrence along the “plan”).
- Removes code to maintain (`maintenance_plan`).

Cons:
- Not all plan requests are created (only for users who already used it previously, in which case an OCA module could be created that by means of a button on the requests allows to create all future requests avoiding/confirming that on completion it will not create the next one).
- Migration scripts in OpenUpgrade to adapt plans to linked stock requests and corresponding recurrence fields. Possible discussion if future uninitiated requests need to be deleted (to avoid confusion) (as they will be created when the next one is finalized).

Having clarified all this, **I am of the opinion that the `maintenance_plan` module should be removed** (having analyzed this before migrating it) and focus it towards core usage, applying the corresponding changes in Openupgrade at https://github.com/OCA/OpenUpgrade/pull/4654.

Ping @pedrobaeza

pedrobaeza:
OK then on stop using this module. Let's do the OU thing, and we can drop in this module a deprecated message in the README.

PR #423, opened 2024-09-09
[17.0][FIX] maintenance_plan: Update context and domain filters to use active_id

hafizhlf:
Refactored the context and domain in maintenance plan views to use `active_id` instead of `id`.

This pull request addresses an issue where the id field was not found in the context and domain of the maintenance plan views.


giarve (approved):
I could reproduce outside runbot on a 17.0

victoralmau:
Oops, looks like this is the same as https://github.com/OCA/maintenance/pull/437, sorry I didn't see it before. 
Ping @pedrobaeza

pedrobaeza:
Yeah, sorry, the other is already merging. It seems nobody checks the open PRs before proposing... Anyway, the main point here is to fix the issue. Closing then.

PR #424, opened 2024-09-13
[16.0][IMP] maintenance_project: Remove the Project > Administrator permission dependency

victoralmau:
Changes done:
- [x] Remove the Project > Administrator permission dependency
- [x] Change `create_project_from_equipment` fields from equipments to '_Create project_' button
- [x] Add `groups="project.group_project_user"` to project/task fields
- [x] Add `groups="maintenance.group_equipment_manager"` to maintenance fields

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT50829


victoralmau:
I have found one additional controversial thing, the project and task fields do not have groups set, i.e. a user without permissions in Projects can see those fields (even if they don't have a permissions error). Is it consistent that they can see them?

carlos-lopez-tecnativa (commented):
Functionally tested, but I have a question: isn't it necessary to have a migration script to unlink the group in existing databases?

maintenance_project/views/maintenance_equipment_views.xml
```diff
@@ -40,10 +40,13 @@
                 <field
                     name="create_project_from_equipment"
                     attrs="{'invisible': [('id', '!=', False)]}"
+                    groups="project.group_project_manager"
                 />
+                <field name="project_id" invisible="1" />
```
dalonsod, line 43: Not sure if this invisible field is needed. If it's because `project_id` is required by `preventive_default_task_id`, this field should be also restricted to `project.group_project_manager` as well.

dalonsod (changes requested):
See comment. And also, should access to project (and task) fields be restricted in the other views, not only for the `form` one, and for `maintenance.request` views too?

dalonsod:
> I have found one additional controversial thing, the project and task fields do not have groups set, i.e. a user without permissions in Projects can see those fields (even if they don't have a permissions error). Is it consistent that they can see them?

@victoralmau looked at this, `project` addon adds a read ACL for every internal user called "project.project on partners" (and the same for tasks)... but task access in partner view is finally restricted to project users, it's something strange IMO. In fact, restricting access to projects (and tasks) in general for non-project users seems to be a good practice, I agree with it. But I've also looked at other OCA typical addon that adds project/task management (`helpdesk_mgmt_project`) and has the same behavior than current implementation of `maintenance_project` (there are no access restrictions!).

I should add such access restriction (not consistent, as you said), but not sure if this change could be controversial as well because a current functionality should be removed.

pedrobaeza:
@dalonsod real access is restricted by record rules, not ACLs.

maintenance_project/views/maintenance_equipment_views.xml
```diff
@@ -40,10 +40,13 @@
                 <field
                     name="create_project_from_equipment"
```
pedrobaeza: Not sure to understand the UX of this check... So you mark this check, and when you write, the project is created? Why not having a button for that?

victoralmau: I totally agree, I do not like this field and it serves only for that purpose (to create a project when the maintenance.equipment is created if this check is checked).

I will make the corresponding change to simplify this behavior.

victoralmau: Removed field and added a button.

victoralmau:
Added extra changes and updated the PR description.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

maintenance_project/views/maintenance_equipment_views.xml
dalonsod: Review refactoring, with this new approach "create project from equipment" flow has been changed:
- Before, the form check enabled creating and initially assigning a new project based on new equipment, only when an equipment is created. Later, the only option was replacing project for other (or leave it empty), never making project clones.
- After, new project based on equipment is available once equipment is created (two steps), and only when project is empty. Indeed, every time equipment project is empty you can create a project, so you could theoretically create more than one with exactly the same name (and that could be confusing).

Not sure how to make new behavior closer to previous one, and still replacing "instrumental" field by an action button (and that's a good idea, agree with it).

pedrobaeza: I don't understand you, David

victoralmau: I honestly don't understand the issue with the current change, but perhaps it would be solved changing `readonly` attrs in the `project_id` field if it is already set and the team is created `([[('project_id', '!=', False),('id', '!=', False)])`. 

With this approach:
- Creating the equipment: You will be able to set a project from the existing ones.
- Equipment is already created: You can use the “_Create project_” button (if not already set) or set a project. It will not be possible to “leave empty” the `project_id` field by UI.

What do you think?

dalonsod: Thanks @victoralmau , I'm sorry if my explanations were not clear enough...

I was only noting that with previous implementation creating & assigning a custom project was a task that could only be achieved during equipment creation, and now it's the opposite, such action could only be done after equipment creation. And I could add that in previous implementation the behavior was creating the project by default. For those environments that that was the desired behavior (custom project for every equipment), the functional instructions for manager users will change.

In the other hand, with this change, assigning a project created from the equipment once equipment it's created becomes possible, and that's a truly functionally improvement. So I have no arguments to not approve this.

Finally, I wouldn't recommend your last suggested change, because once project is filled, it seems that it won't be possible to be changed anymore, if I'm not wrong.

dalonsod (approved):

victoralmau:
IMO this is now ready to merge

PR #434, opened 2024-10-23
[16.0][FIX] maintenance_request_sequence: add create_multi

kluna1998:
Enable batch creation of records by applying the @model_create_multi decorator, improving performance when creating multiple records at once.


etobella (approved):
/ocabot merge patch

PR #437, opened 2024-11-28
[17.0][FIX] maintenance_plan: fix the use of id in act_window

vincent-hatakeyama:
Without this fix, the smart button using the act_window fails with an error.


pedrobaeza (approved):
Let's merge this one in favor of #439, as it's more complete.

/ocabot merge patch

PR #444, opened 2025-01-14
[17.0][IMP] base_maintenance: Fine-tuning of the form view

carlos-lopez-tecnativa:
With this commit https://github.com/odoo/odoo/commit/3bd841086d92246e55771785bcb69501af712ab6 the maintenance request form view now has an independent page. This commit adapts the view to align with the new page structure.

Before
image

After
image

@Tecnativa @pedrobaeza @victoralmau could you please review this

### DIFF
--- a/base_maintenance/views/maintenance_request_views.xml
+++ b/base_maintenance/views/maintenance_request_views.xml
@@ -13,13 +13,9 @@
                     <!--Empty, to be inherited by other modules.-->
                 </div>
             </xpath>
-            <field name="description" position="replace">
-                <notebook colspan="2">
-                    <page name="description_page" string="Description">
-                        <field name="description" />
-                    </page>
-                </notebook>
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
I don't think we should change the name, but instead, adapt the modules using that name as selector.

carlos-lopez-tecnativa:
> I don't think we should change the name, but instead, adapt the modules using that name as selector.

In the form view, the new page only includes a `string` attribute and lacks a `name` attribute. This is why I added the name attribute to the page. Please let me know if I missed anything.

https://github.com/odoo/odoo/blob/17.0/addons/maintenance/views/maintenance_views.xml#L137

victoralmau (approved):

pedrobaeza:
OK

/ocabot merge minor

PR #452, opened 2025-01-27
[18.0][MIG] base_maintenance: Migration to 18.0

BhaveshHeliconia:


carlos-lopez-tecnativa (approved):
LGTM
TT54649
@Tecnativa @pedrobaeza @victoralmau could you please review this.

base_maintenance/i18n/base_maintenance.pot
```diff
@@ -4,7 +4,7 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 16.0\n"
+"Project-Id-Version: Odoo Server 17.0\n"
```
pedrobaeza: Incorrect version. If edited, better to put the correct one.

base_maintenance/tests/test_base_maintenance.py
```diff
@@ -0,0 +1,28 @@
...
+from odoo.tests import TransactionCase
+
+
+class TestMaintenanceTeam(TransactionCase):
+    def setUp(cls):
+        super().setUp()
+        # Setup for creating user and maintenance team records for tests
+        cls.user = cls.env.ref("base.user_admin")
+
+    def test_create_maintenance_team(self):
```
pedrobaeza: This test is not adding value, as it's testing ORM features, not this module features. You can print the report as test of not crashing during its rendering, which adds value.

base_maintenance/views/maintenance_request_views.xml
```diff
@@ -13,13 +13,9 @@
...
                 </div>
             </xpath>
-            <field name="description" position="replace">
-                <notebook colspan="2">
-                    <page name="description_page" string="Description">
-                        <field name="description" />
-                    </page>
-                </notebook>
-            </field>
+            <xpath expr="//field[@name='description']/.." position="attributes">
```
pedrobaeza, line 16: This should be cherry-picked from e5894716102fbeab02281b2ff3a2a79de1904f4e before the migration commit instead of putting it here as if it was part of the migration.

pedrobaeza (changes requested):

base_maintenance/views/maintenance_request_views.xml
BhaveshHeliconia: @pedrobaeza , I cherry-picked the changes and fixed the issues. Could you please review it again?

pedrobaeza (approved):
/ocabot merge nobump

PR #457, opened 2025-01-27
[18.0][MIG] maintenance_plan: Migration to 18.0

BhaveshHeliconia:
Depends on

- https://github.com/OCA/maintenance/pull/452


victoralmau:
This module is not required, even in v17 https://github.com/OCA/maintenance/pull/414#issuecomment-2668929617

PR #468, opened 2025-02-06
[17.0][MIG] maintenance_plan_activity: Migration to 17.0

ArnauCForgeFlow:
Standard migration to 17.0


victoralmau:
IMO this module is not required in v17, neither are the plans https://github.com/OCA/maintenance/pull/414#issuecomment-2668929617 all this is something that can already be done in core.

https://github.com/OCA/OpenUpgrade/pull/4752

PR #474, opened 2025-02-27
[13.0][ADD] maintenance_request_purchase_order_type

dalonsod:
This module enables only those Purchase Orders for selected Order Types to be linked to Maintenance Requests.

This could be useful in order to isolate whose POs that should be used with Maintenance Requests.


dalonsod:
@etobella please review, if addon seems to be useful... when merged, I'll port it to later versions, I promise

PR #481, opened 2025-04-03
[IMP] maintenance_request_purchase: Allow to open the form of related Purchase Orders

etobella:
With this change, we can open the form of the PO. 

It is interesting if we have attachments and messages on the PO (not viewed otherwise)

image


luisDIXMIT (approved):
Tested on runboat, LGTM!

kluna1998 (approved):
Functional test. LGTM 👍🏽

PR #486, opened 2025-04-09
[18.0][MIG] maintenance equipment tags: Migration to 18.0

tokiniainaLANJANOMENA:


BhaveshHeliconia (approved):

marcelsavegnago (changes requested):
It seems you may have skipped some steps in the migration process. Could you please share the commit history so we can verify what was done?

PR #492, opened 2025-05-22
[16.0][IMP] maintenance_request_purchase: add total_purchase_amount in maintenance.request

kluna1998:
add total_purchase_amount in maintenance.request so it can be used in the pivot view

image

### DIFF
--- a/maintenance_request_purchase/models/maintenance_request.py
+++ b/maintenance_request_purchase/models/maintenance_request.py
@@ -23,6 +23,42 @@ class MaintenanceRequest(models.Model):
         groups="purchase.group_purchase_user",
     )
 
+    total_purchase_amount = fields.Monetary(
+        compute="_compute_total_purchase_amount",
+        store=True,
+        groups="purchase.group_purchase_user",
+        currency_field="currency_id",
+    )
+
+    currency_id = fields.Many2one(
+        related="company_id.currency_id",
+        store=True,
+        readonly=True,
+    )
+
+    @api.depends(
+        "purchase_order_ids.amount_total",
+        "purchase_order_ids.currency_id",
+        "company_id.currency_id",
+        "purchase_order_ids.state",
+    )
+    def _compute_total_purchase_amount(self):
+        date = self.env.context.get("actual_date") or fields.Date.today()
+        for record in self:
+            company_currency = record.company_id.currency_id
+            total = sum(
+                po.currency_id._convert(
+                    po.amount_total,
+                    company_currency,
+                    record.company_id,
+                    date,
+                )
+                for po in record.purchase_order_ids.filtered(
+                    lambda po: po.state in ("purchase", "done")
+                )
+            )
+            record.total_purchase_amount = total
+
     @api.depends("purchase_order_ids")
     def _compute_purchases_count(self):
         for record in self:
--- a/maintenance_request_purchase/tests/test_maintenance_purchase.py
+++ b/maintenance_request_purchase/tests/test_maintenance_purchase.py
@@ -22,6 +22,12 @@ class TestMaintenancePurchase(TransactionCase):
                 "date_planned": "2017-02-11 22:00:00",
             }
         )
+        cls.po_2 = cls.env["purchase.order"].create(
+            {
+                "partner_id": cls.supplier.id,
+                "date_planned": "2017-02-11 22:00:00",
+            }
+        )
 
     def test_maintenance_purchase(self):
         self.assertEqual(self.request_1.purchases_count, 0)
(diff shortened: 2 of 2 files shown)


maintenance_request_purchase/models/maintenance_request.py
```diff
@@ -23,6 +23,28 @@ class MaintenanceRequest(models.Model):
...
 
+    total_purchase_amount = fields.Monetary(
+        compute="_compute_total_purchase_amount",
+        store=True,
+        groups="purchase.group_purchase_user",
+    )
+
+    currency_id = fields.Many2one(
+        "res.currency", string="Currency", compute="_compute_currency", store=True
+    )
```
dalonsod: Could you simply try this instead of a computed method?
```suggestion
    currency_id = fields.Many2one(
        related="company_id.currency_id", store=True
    )
```

dalonsod (changes requested):
See comment. Anyway, total purchase amount calculation could be inconsistent, before adding it a proper currency conversion should be performed, not all purchase orders currency should match company currency

kluna1998:
Thanks for the feedback!
You're absolutely right — summing up purchase amounts directly without converting currencies could lead to inconsistencies. I've updated the implementation to ensure all amounts are converted to the maintenance request's company currency using _convert.
Let me know if you see any other potential issues!  @dalonsod

maintenance_request_purchase/models/maintenance_request.py
kluna1998: done 👍🏽

dalonsod:
thanks @kluna1998 I only miss adding a little test for this new field, is it possible?

maintenance_request_purchase/tests/test_maintenance_purchase.py
```diff
@@ -36,3 +36,44 @@ def test_maintenance_purchase(self):
...
+                        {
+                            "product_id": self.env.ref("product.product_product_4").id,
+                            "product_qty": 5,
+                        },
+                    )
+                ]
+            }
+        )
+        self.request_1.purchase_order_ids |= self.po_2
+        self.assertEqual(self.request_1.total_purchase_amount, 7500)
```
dalonsod, line 86: Data for this test is weak, because product data comes from demo products, that could be removed from a version to another.

Anyway, if you use the demo product, at this line you should retrieve product price for the assert, not assuming `500.0`, that it could be changed. Or even better, creating PO lines with a price, so you won't depend on demo product price.

kluna1998: Instead of assuming the unit price, I’ve now explicitly set the price_unit directly on the purchase order lines to make the test independent from the product's price definition.
Sorry for the delay. I totally forgot about this PR. Thanks again for the feedback! @dalonsod

dalonsod (approved):
LGTM

marcelsavegnago (approved):
LGTM

dalonsod:
@etobella could you merge this? Thanks!

maintenance_request_purchase/models/maintenance_request.py
```diff
@@ -23,6 +23,39 @@ class MaintenanceRequest(models.Model):
...
+        for record in self:
+            company_currency = record.company_id.currency_id
+            total = sum(
+                po.currency_id._convert(
+                    po.amount_total,
+                    company_currency,
+                    record.company_id,
+                    date,
+                )
+                for po in record.purchase_order_ids
```
etobella: Shouldn't we remove cancelled purchase orders?

kluna1998: That makes sense! Would it also be appropriate to exclude purchase orders in the draft and to_approve states, so that only confirmed or completed ones are taken into account? @etobella

etobella: Well, the right approachd would be to use confirmed  and blocked POs the to_approve is related to base tier validation and shouldn't be handled here.

maintenance_request_purchase/models/maintenance_request.py
```diff
@@ -23,6 +23,41 @@ class MaintenanceRequest(models.Model):
...
+
+    currency_id = fields.Many2one(
+        related="company_id.currency_id",
+        store=True,
+        readonly=True,
+    )
+
+    @api.depends(
+        "purchase_order_ids.amount_total",
+        "purchase_order_ids.currency_id",
```
dalonsod, line 41: Missing new dependency:
```suggestion
        "purchase_order_ids.currency_id",
        "purchase_order_ids.state",
```

etobella (approved):
/ocabot merge minor

PR #496, opened 2025-07-01
[16.0][IMP] maintenance_security: do not grant HR maintenance groups

giarve:
Odoo grants HR users maintenance permissions by default on hr_maintenance

No need or requirement to do this automatically... because equipments could be not related to HR at all.

Better to use roles for this purpose.

### DIFF
--- a/maintenance_security/README.rst
+++ b/maintenance_security/README.rst
@@ -33,6 +33,8 @@ This module limits the Maintenance menu to users with any of the maintenance gro
 It also removes the assignation mail when a user with no permission is assigned to an
 equipment, as if not, the mail will contain a link to it.
 
+Additionally, it will remove the inheritance of equipment manager maintenance group by HR users (hr.group_hr_user) to further decouple the modules. The group can still be manually assigned to the user.
+
 **Table of contents**
 
 .. contents::
--- a/maintenance_security/readme/DESCRIPTION.rst
+++ b/maintenance_security/readme/DESCRIPTION.rst
@@ -2,3 +2,5 @@ This module limits the Maintenance menu to users with any of the maintenance gro
 
 It also removes the assignation mail when a user with no permission is assigned to an
 equipment, as if not, the mail will contain a link to it.
+
+Additionally, it will remove the inheritance of equipment manager maintenance group by HR users (hr.group_hr_user) to further decouple the modules. The group can still be manually assigned to the user.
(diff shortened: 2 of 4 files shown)


victoralmau:
Ping @pedrobaeza what do you think about adding this here to avoid what odoo does in https://github.com/odoo/odoo/blob/a800803f8108d81548919e494e35a1679beb4622/addons/hr_maintenance/security/equipment.xml#L5 ?

pedrobaeza:
It's a breaking change, but it seems logical. It should be documented in the README though.

giarve:
In my defense, this inheritance was added in 2016 probably to solve a specific use case during an ongoing odoo implementation to close a specific customer ticket. Otherwise, it does not make any sense because HR does not need to be related to maintenance at all. I suppose if they assign equipment to employees, OK, but for sure it is not always the case and better to have it more modular.

victoralmau (approved):

maintenance_security/README.rst
```diff
@@ -33,6 +33,9 @@ This module limits the Maintenance menu to users with any of the maintenance gro
 It also removes the assignation mail when a user with no permission is assigned to an
 equipment, as if not, the mail will contain a link to it.
 
+Additionally, it will remove the inheritance of equipment manager maintenance group by HR users (hr.group_hr_user) to further decouple the modules.
+The group can still be manually assigned to the user. For this kind of automated behaviour is safer to use roles (check the other OCA module).
```
pedrobaeza: I prefer to not have the reference to the roles, as it's unrelated. You can assign the same both groups without roles.

giarve: Yep. Basically I wrote it because roles are the best substitute in these cases for inheritance (in my opinion).

I removed this phrase.

pedrobaeza:
One more thing, sorry: you have to adhere to the commit message guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

pedrobaeza (approved):
/ocabot merge major

victoralmau:
Can you make the changes in v18 as well?

victoralmau:
I don't think this change was appropriate, because if `maintenance_security` is installed and `hr` is not installed, an error will occur because `maintenance_security` does not depend on `hr`.

In my opinion, the appropriate solution would be to create an `hr_maintenance_security` module that depends on `hr_maintenance` and `maintenance_security` to add only these changes.

What do you think?

Related to https://github.com/OCA/maintenance/pull/526

giarve:
Yes, can you revert this commit somehow?

then we merge this

https://github.com/OCA/maintenance/pull/528

PR #501, opened 2025-07-24
[16.0][FIX] maintenance_timesheet: Define the correct rule to show only what is related to maintenance_request_id

victoralmau:
Define the correct rule to show only what is related to `maintenance_request_id`

Locked by:
- [x] `project_timesheet_time_control`: https://github.com/OCA/project/pull/1547

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT56859

### DIFF
--- a/maintenance_timesheet/security/maintenance_timesheet_security.xml
+++ b/maintenance_timesheet/security/maintenance_timesheet_security.xml
@@ -6,11 +6,8 @@
         >Users are allowed to access timesheets related to a followed request</field>
         <field name="model_id" ref="model_account_analytic_line" />
         <field name="domain_force">[
-                '|',
-                    ('maintenance_request_id', '=', False),
-                        '|',
-                            ('maintenance_request_id.message_partner_ids', 'in', [user.partner_id.id]),
-                            ('maintenance_request_id.user_id.id', '=', user.id)
+                ('maintenance_request_id.message_partner_ids', 'in', [user.partner_id.id]),
+                ('maintenance_request_id.user_id.id', '=', user.id)
             ]
         </field>
         <field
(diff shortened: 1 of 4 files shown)


pilarvargas-tecnativa (approved):

pedrobaeza (commented):
The change is not correct as is. I don't say it was correct, but the solution isn't to change `=` to `!=`. That way, you give access to all the timesheets related to maintenance. Maybe you need to just remove that part?

victoralmau:
Right now, I don't know exactly what is not being taken into account.

**Before**
With the existing rule, you had access to **everything** that was not related to maintenance_request (`maintenance_request=False`). That is definitely a total mistake (or at least it is for me).

**After**
The existing rule is intended to be related only to maintenance requests, which is why the rule only takes into account what is related to maintenance requests (`maintenance_request!=False`) and what is “_linked to its user_”.

So what is the appropriate rule?

dalonsod:
> The change is not correct as is. I don't say it was correct, but the solution isn't to change = to !=. That way, you give access to all the timesheets related to maintenance. Maybe you need to just remove that part?

Sure, the original code seems to be incorrect, you're right 😕

The new one already makes more changes than moving `=` to `!=`, it also removes an extra OR, so it's not only, for the timesheet, required belonging to a maintenance request, but following it as well. I see it now right.

maintenance_timesheet/security/maintenance_timesheet_security.xml
```diff
@@ -6,8 +6,7 @@
         >Users are allowed to access timesheets related to a followed request</field>
         <field name="model_id" ref="model_account_analytic_line" />
         <field name="domain_force">[
-                '|',
-                    ('maintenance_request_id', '=', False),
+                    ('maintenance_request_id', '!=', False),
```
pedrobaeza: You can remove it as it's the same effect.

```suggestion
```

pedrobaeza: And deindent the content.

pedrobaeza (approved):

PR #503, opened 2025-07-24
[15.0][FIX] maintenance_timesheet: Define the correct rule to show only what is related to maintenance_request_id

victoralmau:
Backport from 16.0: https://github.com/OCA/maintenance/pull/501

Define the correct rule to show only what is related to `maintenance_request_id`

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT56859


pedrobaeza (approved):
/ocabot merge nobump

PR #505, opened 2025-07-25
[16.0][OU-FIX] maintenance_timesheet: Define the correct path in load_data

victoralmau:
Define the correct path in load_data

Related to https://github.com/OCA/maintenance/pull/501

@Tecnativa


pedrobaeza (approved):
Fast-tracking for fixing the problem

PR #506, opened 2025-07-25
[17.0][OU-FIX] maintenance_timesheet: Define the correct path in load_data

victoralmau:
Define the correct path in load_data

Related to https://github.com/OCA/maintenance/pull/502

@Tecnativa


pedrobaeza (approved):
Bad mistake that can't be checked by usual CIs. Fast-tracking for fixing it.

PR #509, opened 2025-09-01
[IMP] maintenance_project: Add visibility for milestones

etobella:
Milestones comes by default in Projects, but not used here. it was easy to add them.

### DIFF
--- a/maintenance_project/models/maintenance_request.py
+++ b/maintenance_project/models/maintenance_request.py
@@ -9,6 +9,18 @@ class MaintenanceRequest(models.Model):
 
     project_id = fields.Many2one(comodel_name="project.project")
     task_id = fields.Many2one(comodel_name="project.task")
+    milestone_id = fields.Many2one(
+        comodel_name="project.milestone",
+        store=True,
+        compute="_compute_milestone_id",
+        readonly=False,
+    )
+
+    @api.depends("task_id")
+    def _compute_milestone_id(self):
+        for request in self:
+            if request.task_id:
+                request.milestone_id = request.task_id.milestone_id
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/maintenance_project/views/maintenance_request_views.xml
+++ b/maintenance_project/views/maintenance_request_views.xml
@@ -52,7 +52,32 @@
                     context="{'default_project_id': project_id}"
                     groups="project.group_project_user"
                 />
+                <field name="milestone_id" invisible="1" />
+                <t groups="project.group_project_milestone">
+                    <field
+                        name="milestone_id"
+                        domain="[('project_id', '=', project_id)]"
+                        groups="project.group_project_user"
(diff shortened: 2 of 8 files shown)


maintenance_project/models/maintenance_request.py
```diff
@@ -9,6 +9,18 @@ class MaintenanceRequest(models.Model):
 
     project_id = fields.Many2one(comodel_name="project.project")
     task_id = fields.Many2one(comodel_name="project.task")
+    milestone_id = fields.Many2one(
+        comodel_name="project.milestone",
+        store=True,
+        domain="[('project_id', '=', project_id)]",
+        compute="_compute_milestone",
```
dalonsod: Better naming?
```suggestion
        compute="_compute_milestone_id",
```

maintenance_project/views/maintenance_request_views.xml
```diff
@@ -52,7 +52,32 @@
                     context="{'default_project_id': project_id}"
                     groups="project.group_project_user"
                 />
+                <field name="milestone_id" invisible="1" />
+                <t groups="project.group_project_milestone">
+                    <field
+                        name="milestone_id"
+                        domain="[('project_id', '=', project_id)]"
```
dalonsod, line 59: Is this domain required, as already defined at field?

maintenance_project/views/project_milestone.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="arch" type="xml">
+            <button name="action_view_tasks" position="after">
+                <button
+                    name="action_view_maintenance_request"
+                    type="object"
+                    title="View Maintenance Requests"
+                    string="View Maintenance Requests"
+                    class="btn btn-link float-end"
+                    invisible="maintenance_request_count == 0"
+                    groups="maintenance.group_equipment_manager"
```
dalonsod, line 42: Every internal user can access Maintenance Requests by default. Then, I'm not sure if this button should have such restriction, what do you think?

dalonsod (changes requested):
Code and functional review, see comments. And also:
* Pending add some tests for the new python code.
* Add some demo example.

maintenance_project/views/project_milestone.xml
etobella: Well, I applied the same logic that existed on project.project

etobella:
@dalonsod comments attended

dalonsod (approved):
:+1: 

Only last thing, because I'm not sure: after this improvement installation, milestone for existing requests will be initialized with task milestone (if both filled). For certain installations the amount of existing requests could be very high. Could `_compute_milestone_id()` execution make this upgrade installation step very slow and require a migration hook?

etobella:
That is true, I will make a migration script for this it was a job to do, do not worry :wink:

etobella:
Migration script done and tested.

PR #514, opened 2025-10-03
[18.0][MIG] maintenance_plan: Migration to 18.0 

gbrito:


pedrobaeza:
Odoo already includes a way to do recurrent maintenance requests, so this one is less needed. Do you have any reason to keep this one?

ypapouin:
I can read that `maintenance_plan` is deprecated but merged in 17.0 https://github.com/OCA/maintenance/pull/414

Currently some other modules "depends" on it (from 14.0 to 17.0):
- `maintenance_inspection`: https://github.com/OCA/maintenance/tree/14.0/maintenance_inspection
- `maintenance_location`: https://github.com/OCA/maintenance/tree/17.0/maintenance_location
- `maintenance_plan_activity`: https://github.com/OCA/maintenance/tree/16.0/maintenance_plan_activity
- `maintenance_project_plan`: https://github.com/OCA/maintenance/tree/14.0/maintenance_project_plan

For some of them (`maintenance_inspection`,  `maintenance_location`), it would only require to drop references to `maintenance_plan` items to keep them in 18.0.

But that's not the point: the "recurrent" implementation from Odoo is javascript-only ... and its behavior has nothing to do with `maintenance_plan`. I think there has been a misunderstanding on this because the “advanced” features of the `maintenance_plan` module are much closer and more realistic to the way a maintenance workshop operates and this module should be "de-deprecated".

To avoid confusion on the "recurrent" attribute, this field could be hidden when `maintenance_plan` is installed or recomputed (it is already a compute field). Same thing for the calendar view where the `js_class="calendar_with_recurrence"` would be dropped.

I'm ready to do this and continue this migration.

maintenance_plan/views/maintenance_plan_views.xml
```diff
@@ -0,0 +1,204 @@
...
+                        </page>
+                    </notebook>
+                </sheet>
+                <div class="o_attachment_preview" />
+                <!-- Chatter -->
+                <div class="oe_chatter">
+                    <field name="message_follower_ids" groups="base.group_user" />
+                    <field name="activity_ids" />
+                    <field name="message_ids" />
+                </div>
```
ypapouin, line 126: You must use the <chatter> tag

PR #515, opened 2025-10-03
[19.0][MIG] base_maintenance: Migration to 19.0

cparadis-impressfoods:


dalonsod:
Hello @cparadis-impressfoods thanks for the contribution!!

Perhaps you may wait for technical instructions regarded to 19.0 Migration, that are not available yet (https://github.com/OCA/maintainer-tools/wiki/).

At least, I see that commit history was not preserved, and that task is always mandatory for every migration, you can check for other Odoo version migration guidelines.

PR #516, opened 2025-10-13
[18.0][MIG] maintenance_equipment_image: Migration to 18.0

ypapouin:
Big headache for a basic thing.

Demo pictures have been added:
image

My first idea was to use a combination of a wildcard expression `expr=".../*" ` (aka: select all children) with `position="replace"`.
But Odoo is not working like this, node locate only returns the first value (`return nodes[0] if nodes else None`)

```
    <record id="hr_equipment_view_kanban" model="ir.ui.view">
        <field name="model">maintenance.equipment</field>
        <field name="inherit_id" ref="maintenance.hr_equipment_view_kanban" />
        <field name="arch" type="xml">
            <xpath expr="//t[@t-name='card']/*" position="replace" >
                <t t-name="card" class="flex-row">
                    <aside class="o_kanban_aside_full">
                        <field
                            name="image_128"
                            options="{
                                'placeholder': '/base/static/img/company_image.png',
                                'size': [150, 150],
                                'img_class': 'object-fit-cover'
                            }"
                            widget="image"
                        />
                    </aside>
                    <main class="ps-2 ms-1">$0</main>
                </t>
            </xpath>
        </field>
    </record>
```

Since there is no way to use an XPath expression to replace/move all children at once. I had to do it manually for each child.


mamcode (commented):
Hello @ypapouin thanks for this migration, all ok with the module, however I think that the image should be better positioned in the view, I haven't checked in detail but if the maintenance module's kanban view layout allows it, we should do it similarly to how Odoo handles it in the product module. Here you can view the code https://github.com/odoo/odoo/blob/037c4dc20d56aea6351c90a97d3add1dda054f4c/addons/product/views/product_template_views.xml#L123

Image

This is how the view currently looks with the code from this PR
Image

maintenance_equipment_image/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,53 @@
...
+                />
+                <xpath
+                    expr="//t[@t-name='card']/field[@name='serial_no']"
+                    position="move"
+                />
+                <xpath expr="//t[@t-name='card']/footer" position="move" />
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
BhaveshHeliconia, line 49:
```suggestion
<record id="hr_equipment_view_kanban" model="ir.ui.view">
    <field name="model">maintenance.equipment</field>
    <field name="inherit_id" ref="maintenance.hr_equipment_view_kanban" />
    <field name="arch" type="xml">
        <xpath expr="//t[@t-name='card']" position="attributes">
            <attribute name="class">flex-row</attribute>
        </xpath>
        <xpath expr="//t[@t-name='card']/*[1]" position="before">
            <aside class="o_kanban_aside_full">
                <field
                    name="image_128"
                    widget="image"
                    options="{'img_class': 'w-100 object-fit-contain'}"
                    invisible="not image_128"
                />
            </aside>
        </xpath>
    </field>
</record>
```

BhaveshHeliconia (changes requested):
Agreed with @mamcode on the card layout! Adopting core Odoo 18's product pattern fixes the alignment issues and removes the need to manually move fields.

maintenance_equipment_image/views/maintenance_equipment_views.xml
ypapouin: Not working, the `<aside>` tag must be inside the `<t t-name="card">`

ypapouin:
@mamcode, my implementation is exactly the same than the product's one. The `<aside>` tag must be inside the `<t t-name="card">`, like any other implementations. And you need a `<main>` tag before or after that must contains everything else. 

Odoo added a new `equipment_properties` field, exactly one month ago, in the equipment kanban card that breaks the rendering (see [ odoo/32c5ef25](https://github.com/odoo/odoo/commit/32c5ef25)).

Like I wrote before, there is now way to use an XPath expression to replace/move all children at once, and I had to do it manually for each child ... but I thought about another solution: Instead of moving children, I move the entire card with its children but before that, I renamed it so Odoo ignore it and taadaaa.

So, If Odoo add new fields in the future, it should not breaks.

BhaveshHeliconia (approved):

PR #517, opened 2025-10-14
[18.0][MIG] maintenance_team_hierarchy: Migration to 18.0

ypapouin:
Basic migration task.

- Added the parent field in the list view (editable)
- Also added the "View Form" shortcut using `open_form_view=True`

image


maintenance_team_hierarchy/views/maintenance_team.xml
```diff
@@ -2,8 +2,21 @@
...
+                <attribute name="open_form_view">True</attribute>
+            </xpath>
+            <field name="member_ids" position="after">
+                <field name="parent_id" />
+            </field>
+        </field>
+    </record>
     <record model="ir.ui.view" id="maintenance_team_form_view">
-        <field name="name">maintenance.team.form (in maintenance_team_hierarchy)</field>
+        <field name="name">maintenance.team.form@maintenance_team_hierarchy</field>
```
sergiocorato: Why this change?

ypapouin: Just because this is my own way to set view names for our modules and since there is no convention about this, I think its more easy to visualize and to parse.
But I could simply remove it since the xml-id is now always displayed.

maintenance_team_hierarchy/models/maintenance_team.py
```diff
@@ -11,7 +11,7 @@ class MaintenanceTeam(models.Model):
     _inherit = "maintenance.team"
 
     parent_id = fields.Many2one("maintenance.team", ondelete="restrict")
-    parent_path = fields.Char(index=True, unaccent=False)
+    parent_path = fields.Char(index=True)
```
sergiocorato, line 14: `unaccent` key does not exist I see, maybe it's worth to put in the migration wiki?
I cannot see it till now.

sergiocorato (changes requested):
Please squash administrative and other commits as explained in the migration wiki https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

ypapouin:
@sergiocorato , I don't see anything that could be squashed according to the Wiki

sergiocorato:
> @sergiocorato , I don't see anything that could be squashed according to the Wiki

example, these should be merged in the first:

Screenshot(208)

other example, these 3 should be merged in the first too:

Screenshot(210)

ypapouin:
In the Wiki example, only commits with same commit message are squashed and that make sense

maintenance_team_hierarchy/views/maintenance_team.xml
sergiocorato: Migrations shouldn't have these changes, as in this way anyone would change something following his personal opinions.
So please remove them, as unnecessary.

Obviously, if you develop a new module, you can do as you wish.

sergiocorato:
> In the Wiki example, only commits with same commit message are squashed and that make sense

I'm referring to this part for the first example:

Screenshot(211)

sergiocorato:
To this part for the second:

Screenshot(212)

ypapouin:
Ok, so does that mean that , for example, I should squash these 3 commits together ?
(and keep only the message "[MIG] maintenance_team_hierarchy: Migration to 15.0")

image

sergiocorato:
> Ok, so does that mean that , for example, I should squash these 3 commits together ? (and keep only the message "[MIG] maintenance_team_hierarchy: Migration to 15.0")
>

This screen has not enough informations, if you watch here in "commits" section you can see as explained before and as found in migration guidelines, below what I should do:

Screenshot(216)

ypapouin:
@sergiocorato , thank you for taking time to explain all this.
I think that everything is now properly squashed.

sergiocorato (approved):

sergiocorato:
@OCA/maintenance-maintainers could this one be merged?

maintenance_team_hierarchy/views/maintenance_team.xml
```diff
@@ -2,6 +2,19 @@
 <!-- Copyright 2019 Creu Blanca
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
+    <record model="ir.ui.view" id="maintenance_team_tree_view">
+        <field name="name">maintenance.team.tree (in maintenance_team_hierarchy)</field>
+        <field name="model">maintenance.team</field>
+        <field name="inherit_id" ref="maintenance.maintenance_team_view_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//list" position="attributes">
+                <attribute name="open_form_view">True</attribute>
```
etobella, line 11: Why this open_form_view?

tonygalmiche (approved):
Tested on Odoo 18.0: created a parent/child team hierarchy, inline-edited the `parent_id` field in the team list, and confirmed the "View" link opens the full form (needed since the list is now editable — answers the `open_form_view` question raised in review).

Created a maintenance request on a child team and verified it correctly rolls up into the parent team's counters and search filters (`child_of` aggregation).

Everything works as expected.

etobella (approved):
👍

PR #520, opened 2025-10-16
[18.0][MIG] maintenance_plan: Another Migration to 18.0

ypapouin:
Continuation from https://github.com/OCA/maintenance/pull/514, I'm reusing my comment from https://github.com/OCA/maintenance/pull/514#issuecomment-3398078981

You can read that `maintenance_plan` is deprecated but merged in 17.0 https://github.com/OCA/maintenance/pull/414

Currently some other modules "depends" on it (from 14.0 to 17.0):
- `maintenance_inspection`: https://github.com/OCA/maintenance/tree/14.0/maintenance_inspection
- `maintenance_location`: https://github.com/OCA/maintenance/tree/17.0/maintenance_location
- `maintenance_plan_activity`: https://github.com/OCA/maintenance/tree/16.0/maintenance_plan_activity
- `maintenance_project_plan`: https://github.com/OCA/maintenance/tree/14.0/maintenance_project_plan

For some of them (`maintenance_inspection`,  `maintenance_location`), it would only require to drop references to `maintenance_plan` items to keep them in 18.0.

But that's not the point: the "recurrent" implementation from Odoo is javascript-only ... and its behavior has nothing to do with `maintenance_plan`. I think there has been a misunderstanding on this because the “advanced” features of the `maintenance_plan` module are much closer and more realistic to the way a maintenance workshop operates and this module should be reintroduced, that's the point of this PR.

Following are changes:

- To avoid confusion on the "recurrent" attribute, built-in fields are hidden when `maintenance_plan` is installed
- `_compute_recurring_maintenance` is always `False`
- The calendar view javascript attribute `js_class="calendar_with_recurrence"` is removed.
- The `note` field from `maintenance.request` has been dropped (use instructions fields instead)
- The `note` field from `maintenance.plan` has been replaced with the same instructions fields existing in `maintenance_request` since 17.0.

Remaining things to do:

- Migration script to convert `note` field content but how ?
- Create an intermediate wizard for the "Create request from maintenance plan" button to allow selecting a default start hour and selecting valid days of week.
- Backport this PR when merged.


ThiagoMForgeFlow:
I’m not sure if the best way to maintain the functionalities of the child modules is to migrate this module and make Odoo’s standard functionality invisible. Perhaps we should modify the child modules so that instead of inheriting from maintenance_plan, they inherit from base_maintenance, and then refactor their functionalities. What do you think @pedrobaeza ?

etobella:
Well, after some time struggling with the Odoo way and the maintenance plan way I think I can share my POV.

1- Odoo way is simpler, however, it works well only in controlled environments. For example, If I need to make a maintenance plan for reviews every 3 years, Odoo allows me to do it, but it creates some preventive tasks in advance that have no sense IMO.
2- Odoo way requires the creation of the first task. at some point, we added the option of autodetection of equipments. That was much better IMO.
3- Maintenance plan is complicated might be harder to configure
4- Maintenance plan allows me to create tasks in advance at the exact point I want to.

IMO, we should keep both options in the system and make maintenance_plan work in what makes sense, complex environments. It means that we shouldn't hide the odoo way, however, I would recommend to use maintenance plan if possible. About the migrations, we should need to do some stuff to avoid the execution of the migration scripts, but I have an idea of how to do it.

pedrobaeza:
You can explain that on the CONTEXT.md file of this module.

maintenance_plan/models/maintenance_plan.py
```diff
@@ -0,0 +1,272 @@
...
+        """Prepare the context used when evaluating python code
+        :returns: dict -- evaluation context given to safe_eval
+        """
+        return {
+            "datetime": safe_eval.datetime,
+            "dateutil": safe_eval.dateutil,
+            "time": safe_eval.time,
+        }
+
+    def name_get(self):
```
ThiagoMForgeFlow: In v18 this method no longer exists; you should inherit the _compute_display_name method instead.

maintenance_plan/models/maintenance_plan.py
```diff
@@ -0,0 +1,272 @@
...
+        string="Instruction",
+        default="text",
+    )
+    instruction_pdf = fields.Binary("PDF")
+    instruction_google_slide = fields.Char(
+        "Google Slide",
+        help="Paste the url of your Google Slide. "
+        "Make sure the access to the document is public.",
+    )
+    instruction_text = fields.Html("Text")
```
ThiagoMForgeFlow, line 79: To migrate the data that was stored in the note field up to version 17, we need a migration script in the folder migrations/18.0.1.0/pre-migration.py with the following code:
```

from openupgradelib import openupgrade

@openupgrade.migrate()
def migrate(env, version):
    if openupgrade.column_exists(env.cr, "maintenance_plan", "note"):
        openupgrade.rename_fields(
            env,
            [
                (
                    "maintenance.plan",
                    "maintenance_plan",
                    "note",
                    "instruction_text",
                ),
            ],
        )
```

maintenance_plan/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                        <field
+                            name="maintenance_plan_horizon"
+                            string="P. Horizon period"
+                        />
+                        <field name="planning_step" string="P. Horizon step" />
+                    </list>
+                </field>
+            </xpath>
+            <field name="maintenance_team_id" position="after">
+                <field name="maintenance_team_required" invisible="True" />
```
ThiagoMForgeFlow: In v18 it is no longer necessary to add fields with the invisible="true" attribute, you can remove this from the view.

ThiagoMForgeFlow (changes requested):

ThiagoMForgeFlow:
> Well, after some time struggling with the Odoo way and the maintenance plan way I think I can share my POV.
> 
> 1- Odoo way is simpler, however, it works well only in controlled environments. For example, If I need to make a maintenance plan for reviews every 3 years, Odoo allows me to do it, but it creates some preventive tasks in advance that have no sense IMO. 2- Odoo way requires the creation of the first task. at some point, we added the option of autodetection of equipments. That was much better IMO. 3- Maintenance plan is complicated might be harder to configure 4- Maintenance plan allows me to create tasks in advance at the exact point I want to.
> 
> IMO, we should keep both options in the system and make maintenance_plan work in what makes sense, complex environments. It means that we shouldn't hide the odoo way, however, I would recommend to use maintenance plan if possible. About the migrations, we should need to do some stuff to avoid the execution of the migration scripts, but I have an idea of how to do it.

Hi @etobella, have you been able to spend some time on the proposal for making this module’s functionalities coexist with the ones provided by Odoo?

etobella:
@ThiagoMForgeFlow I didn't, there were no more answer on the topic, however, it might have sense IMO to restore it

ThiagoMForgeFlow:
I agree that the module should be kept in version 18 @etobella. But I don’t think Odoo’s standard functionality should be hidden; perhaps you could add a setting to decide whether it is visible or not at the equipment or company level. What do you think @ypapouin ?

etobella:
I do agree with your approach

ypapouin:
I agree that the fields could easily be hidden using an option (company/equipment), but the biggest problem comes from the calendar. It uses a specific JavaScript implementation, and in fact, Odoo's entire recurring implementation is done in this calendar.

Honestly, I don't know how this option could be implemented to disable "maintenance_plan" inherited views for a specific company/equipment. (I don't really see the point of having an option to hide or show this feature, since once you install the “maintenance_plan” module, you accept that Odoo's built-in recurrent features will not be visible.)

etobella:
I think both approaches have their own target, and might even have sense to have them together in some specific environments.

I think we shouldn't remove odoo maintenance behavior by default. If that is a problem, we can have a second module for that.

ypapouin:
> we can have a second module for that.

I'm Ok with that, I will extract every "hide/disable" view items to a specific module and add a `res.config.settings` to install it.
Which name for a such module ? `maintenance_plan_only` ?

etobella:
Yes, it seems correct to me.

ypapouin:
As requested, everything related to hide Odoo builtin recurrent maintenance has been moved to a new module.
The last commit adds the `maintenance_plan_only` module that can be enabled using a setting:
image

maintenance_plan/views/maintenance_equipment_views.xml
```diff
@@ -0,0 +1,68 @@
...
+                >
+                    <field
+                        string="Plans"
+                        name="search_maintenance_plan_count"
+                        widget="statinfo"
+                    />
+                </button>
+            </div>
+            <xpath expr="//group[@name='statistics']" position="attributes">
+                <attribute name="invisible">True</attribute>
```
cvinh, line 23: Does anybody remember why the statistic data have been hidden ?

ThiagoMForgeFlow: Hi @cvinh, after checking the commit history I can see that it was originally added in version 12, replacing the group maintenance. Then, in the migrations for versions 13 and 17, the xpath was adjusted to match each Odoo version. However, looking at Odoo version 18, I’m not sure it still makes sense to hide these fields.

original implemenatation: [12_imp](https://github.com/OCA/maintenance/commit/10cd585153bf9bda8807f93e1332982157124ae9#diff-c5ed3c8212de850d39e065143984c883ee5e41983ae4ac2ff9a20d17c18c3df2R18)

Migration commit 13: [13_mig](https://github.com/OCA/maintenance/commit/6a5739d3a822abc734b323db29ce4690fedeace8#diff-c5ed3c8212de850d39e065143984c883ee5e41983ae4ac2ff9a20d17c18c3df2R18)

Migration commit 17: [17_mig](https://github.com/OCA/maintenance/commit/9f99863b12f257cc9ab2f4f34aa3f7cafd4eb0d5#diff-c5ed3c8212de850d39e065143984c883ee5e41983ae4ac2ff9a20d17c18c3df2R23)

v18 fields:

ThiagoMForgeFlow (approved):

ThiagoMForgeFlow:
Hi @etobella and @pedrobaeza What is your final opinion?

pedrobaeza:
I let @etobella to give his opinion, as we are going to the native feature.

etobella (approved):
Let's proceed with it

/ocabot merge nobump

PR #523, opened 2025-10-21
[18.0][MIG] maintenance_location: Migration to 18.0

ypapouin:
Depends on:

- https://github.com/OCA/maintenance/pull/520


maintenance_location/models/maintenance_location.py
```diff
@@ -0,0 +1,52 @@
...
+                location.complete_name = (
+                    f"{location.parent_id.complete_name} / {location.name}"
+                )
+            else:
+                location.complete_name = location.name
+
+    @api.constrains("parent_id")
+    def _check_category_recursion(self):
+        if self._has_cycle():
+            raise ValidationError(_("Error ! You cannot create recursive Locations."))
```
Quan-nhm: should change from `_` to `self.env._`
it belong to the migration guides of version 18.0
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

> You can replace the call to translate a string () with self.env. for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

You should also remove the import `_`

Quan-nhm (changes requested):

maintenance_location/views/maintenance_equipment.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2019 Creu Blanca
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record model="ir.ui.view" id="hr_equipment_view_form">
+        <field name="name">maintenance.equipment.form@maintenance_location</field>
```
dalonsod, line 6: Are these changes at `view` names required?

maintenance_location/views/maintenance_location.xml
```diff
@@ -0,0 +1,61 @@
...
+                <field name="complete_name" />
+            </list>
+        </field>
+    </record>
+    <record model="ir.actions.act_window" id="maintenance_location_act_window">
+        <field name="name">Maintenance Location</field>
+        <field name="res_model">maintenance.location</field>
+        <field name="view_mode">list,form</field>
+        <field name="domain">[]</field>
+        <field name="context">{}</field>
```
dalonsod, line 53: You could consider adding a `path` for this action

dalonsod (changes requested):
Only code review, minor comments

cgarcia-solvos:
Hello @ypapouin could cherry-pick https://github.com/OCA/maintenance/pull/546 ?

maintenance_location/views/maintenance_location.xml
ypapouin: Path added

maintenance_location/views/maintenance_equipment.xml
ypapouin: All view names changes are reverted

ypapouin:
@cgarcia-solvos migration re-done on current 17.0 head

dalonsod (approved):
👍

BhaveshHeliconia (approved):
Code review LGTM!

maintenance_location/models/maintenance_request.py
```diff
@@ -0,0 +1,21 @@
...
+
+    location_id = fields.Many2one(
+        "maintenance.location",
+        compute="_compute_equipment_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("equipment_id")
+    def _compute_equipment_id(self):
```
cgarcia-solvos: "_compute_equipment_id" actually computes the "location_id". Could you rename it to "_compute_location_id"?


```suggestion
    def _compute_location_id(self):
```

cgarcia-solvos (commented):
Code Review. LGTM! Only minor comment

cgarcia-solvos (approved):

dalonsod:
@yvaucher is this ready to be merged? Thanks!

maintenance_location/models/maintenance_request.py
```diff
@@ -9,13 +9,15 @@ class MaintenanceRequest(models.Model):
...
         "maintenance.location",
-        compute="_compute_equipment_id",
+        compute="_compute_location_id",
         store=True,
         readonly=False,
     )
 
-    @api.depends("equipment_id")
-    def _compute_equipment_id(self):
+    @api.depends("equipment_id", "equipment_id.location_id")
```
etobella: No, it was done this way for a reason. Imagine I have an equipment in location A, I do a task, and it will be assigned to A. A few months after the requests, I move the equipment from A to B, the request should stay as it was, however, with this change, it will move from A to B.

etobella (changes requested):

maintenance_location/models/maintenance_request.py
ypapouin: I understand your point, but since changing the location of an equipement should change all open requests, an alternative implementation could be:

```
    @api.depends(
        "equipment_id", "equipment_id.location_id", "stage_id", "stage_id.done"
    )
    def _compute_location_id(self):
        for record in self:
            if record.equipment_id and record.equipment_id.location_id:
                if record.stage_id and record.stage_id.done:
                    # if the request is done, we want keep existing location
                    pass
                else:
                    record.location_id = record.equipment_id.location_id
            else:
                record.location_id = False
```

I know that setting a value for a computed field was mandatory since Odoo 14, but it does not apply for stored values:

image

etobella: :+1:

ypapouin: Added with updated tests

etobella (approved):

PR #521, opened 2025-10-21
[18.0][MIG] maintenance_inspection: Migration to 18.0

ypapouin:
Depends on 

- https://github.com/OCA/maintenance/pull/520

PR #524, opened 2025-11-05
[17.0][MIG] maintenance_inspection

edescalona:
@BinhexTeam

Based on

- [ ] #521


edescalona:
Hi, @ypapouin @etobella @mymage, if you could do a review, thank you.

ypapouin (changes requested):
You cannot re-use my PR as is. First, because I believe that action path has been introduced in 18.0, and secondly because Odoo 17.0 is using the legacy Kanban view.
You have to remove these two commits:
- [MIG] maintenance_inspection: Kanban view updated
- [IMP] maintenance_plan: Path added to action

image

edescalona:
ping @ypapouin

maintenance_inspection/models/maintenance_inspection_line.py
```diff
@@ -0,0 +1,30 @@
...
+
+    request_id = fields.Many2one(
+        "maintenance.request", required=True, ondelete="cascade"
+    )
+    item_id = fields.Many2one("maintenance.inspection.item", required=True)
+    instruction = fields.Text(related="item_id.instruction")
+    result = fields.Selection(
+        [("todo", "Todo"), ("success", "Success"), ("failure", "Failure")],
+        default="todo",
+        readonly=True,
```
rrebollo, line 19:
```suggestion
```
Related fields are read only by default

maintenance_inspection/models/maintenance_plan.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MaintenancePlan(models.Model):
+    _inherit = "maintenance.plan"
+
+    inspection_item_ids = fields.Many2many("maintenance.inspection.item")
```
rrebollo, line 10: You could set relation attribute to prevent collisions or use a nice name instead of the ORM generated one.

maintenance_inspection/models/maintenance_request.py
```diff
@@ -0,0 +1,36 @@
...
+
+class MaintenanceRequest(models.Model):
+    _inherit = "maintenance.request"
+
+    has_inspection = fields.Boolean()
+    inspection_line_ids = fields.One2many(
+        "maintenance.inspection.line", inverse_name="request_id"
+    )
+    inspection_closed_at = fields.Datetime(readonly=True, copy=False)
+    inspection_closed_by = fields.Many2one("res.users", readonly=True, copy=False)
```
rrebollo: You might consider this suggestion from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

`If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.`

maintenance_inspection/models/maintenance_request.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def set_inspection(self):
+        self.write({"has_inspection": True})
```
rrebollo:
```suggestion
        self.has_inspection = True
```

maintenance_inspection/models/maintenance_request.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def finish_inspection(self):
+        self.write(
```
rrebollo:
```suggestion
        self.update(
```

maintenance_inspection/models/maintenance_request.py
```diff
@@ -0,0 +1,36 @@
...
+            {
+                "inspection_closed_at": fields.Datetime.now(),
+                "inspection_closed_by": self.env.user.id,
+            }
+        )
+
+    def reset_equipment_request(self):
+        res = super().reset_equipment_request()
+        self.write(
```
rrebollo:
```suggestion
        self.update(
```

maintenance_inspection/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+Add the concept of inspection items inside Maintenance requests.
+
+Will only be usable insidea preventive requests.
```
rrebollo:
```suggestion
Will only be usable inside a preventive requests.
```

rrebollo (commented):
Please consider my suggestions. Also, I think the last commit should be squashed.

maintenance_inspection/models/maintenance_inspection_line.py
etobella: yes, but this is not a related field 🤔

maintenance_inspection/models/maintenance_request.py
etobella: It would require the self.ensure_one, I would prefer to keep a self.update better

maintenance_inspection/models/maintenance_inspection_line.py
rrebollo: Sorry, I did a mess. I was thinking on the suggestion from OCA migration guidelines to 17.0 that I mention below.

maintenance_inspection/models/maintenance_request.py
rrebollo: @edescalona @etobella It's just a suggestion, but as a matter of fact it won't require `ensure_one()`, ORM now handle this well with recordsets too. I recently learned this while doing a review over here..

```
env: <odoo.api.Environment object at 0x7ae370b34fa0>
odoo: <module 'odoo' from '/opt/odoo/custom/src/odoo/odoo/__init__.py'>
openerp: <module 'odoo' from '/opt/odoo/custom/src/odoo/odoo/__init__.py'>
self: res.users(1,)
Python 3.10.15 (main, Nov 12 2024, 02:24:06) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(Console)
>>> partners = env['res.partner'].search([])
>>> partners.name = "The same"
>>> env.cr.commit()
>>> exit()
```

image

edescalona:
> Please consider my suggestions. Also, I think the last commit should be squashed.

I think it would be better to leave that revert history as is, don't you think?

rrebollo:
> I think it would be better to leave that revert history as is, don't you think?

Let's see what others think about it

edescalona:
> > I think it would be better to leave that revert history as is, don't you think?
> 
> Let's see what others think about it

I have no problem removing it; I just checked different pull requests and there are commits like that, but I'll proceed to remove it. Thanks.

edescalona:
ping @rrebollo @etobella @ypapouin

maintenance_inspection/models/maintenance_equipment.py
```diff
@@ -0,0 +1,30 @@
...
+                {
+                    "has_inspection": True,
+                    "inspection_line_ids": [
+                        (
+                            0,
+                            0,
+                            {
+                                "item_id": item.id,
+                            },
+                        )
```
rrebollo: I missed this before. What about using `Command`

maintenance_inspection/models/maintenance_inspection_line.py
```diff
@@ -0,0 +1,30 @@
...
+        required=True,
+        copy=False,
+    )
+    result_description = fields.Char()
+    inspection_closed_at = fields.Datetime(related="request_id.inspection_closed_at")
+
+    def action_success(self):
+        self.write({"result": "success"})
```
rrebollo:
```suggestion
        self.update({"result": "success"})
```

maintenance_inspection/models/maintenance_inspection_line.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def action_failure(self):
+        self.write({"result": "failure"})
```
rrebollo:
```suggestion
        self.update({"result": "failure"})
```

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

etobella:
One silly questions, a lot of the changed done in the migration should be in the 18 migration too, like the "readonly" fields or Command changes.

ypapouin:
@edescalona , I previously wrote that you have to remove these two (non-squashed on purpose) commits to be compatible with 17.0:
- [MIG] maintenance_inspection: Kanban view updated
- [IMP] maintenance_plan: Path added to action
But instead of doing it, you have reverted these commits changes directly in your "Migration to 17.0" commit. That's not how it should be done.

@etobella  I can replicate code changes on my 18.0 PR since all requests are relevant.

Maybe it's more "clean" to make changes in the 18.0 PR and rebase this PR after that ?

edescalona:
Hi @ypapouin , well I understand that either way would be fine to make the changes, either with reverts or directly in the migration configuration. I can do it in the most convenient way. @etobella , what do you think about that?

mymage (approved):
Functional OK

etobella:
@edescalona  if some commits were splitted on purpose so you don't need to import it, please remove them.

edescalona:
Hi @ypapouin , did you make the suggested changes in this migration to the 18 version? Can you confirm so I can perform the rebase again? Thanks.

ypapouin:
Hi @edescalona, I'm on it right now

ypapouin:
@edescalona , PR https://github.com/OCA/maintenance/pull/521 is ready to rebase this one

Like before drop these commits:

- [MIG] maintenance_inspection: Kanban view updated
- [IMP] maintenance_plan: Path added to action

The last edit should be to rename `list` to `tree`

edescalona:
Hi @ypapouin, thanks for letting me know.

edescalona:
ping @ypapouin @etobella

PR #526, opened 2025-11-23
[18.0][FIX] maintenance_security: add hr module dependancy

cvinh:
maintenance_security fails at install because it does not depend on hr module


flotho (changes requested):
Hi,
Thanks for the fix,
any chance to change PR and commit names ?

PR #528, opened 2025-11-24
[16.0][IMP] hr_maintenance_security: split into bridge module

giarve:


hr_maintenance_security/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Maintenance Security",
+    "version": "16.0.1.0.0",
+    "category": "Tools",
+    "website": "https://github.com/OCA/maintenance",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["hr_maintenance"],
```
victoralmau:
```suggestion
    "depends": ["hr_maintenance", "maintenance_security"],
```

victoralmau (changes requested):

victoralmau (approved):
Code review OK

Ping @pedrobaeza 

I think this PR could be used to remove unnecessary code from `maintenance_security` that is added in this new module.

pedrobaeza (approved):
But the revert of the other commit is pending, isn't it?

/ocabot merge nobump

PR #533, opened 2025-11-25
[18.0][MIG] maintenance_equipment_meter

cvinh:


maintenance_equipment_meter/models/maintenance_request.py
```diff
@@ -0,0 +1,50 @@
...
+    @api.depends("meter_id.value")
+    def _compute_meter(self):
+        for record in self:
+            record.current_meter = record.meter_id.value or 0.0
+
+    def _inverse_meter(self):
+        for rec in self:
+            if not rec.current_meter and self.meter_id:
+                raise UserError(
+                    _("Emptying the odometer value of an equipment is not allowed.")
```
Quan-nhm: should change from `_` to `self.env._`
it belong to the migration guides of version 18.0
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

> You can replace the call to translate a string (_) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

Quan-nhm (changes requested):

maintenance_equipment_meter/models/maintenance_request.py
cvinh: Done

maintenance_equipment_meter/models/maintenance_request.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2023 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
```
Quan-nhm: Due to we are not use `_` I thin kwe should remove this from the import as well.
It would be nice if you could check the pre-commit and codecov as well.

Quan-nhm:
by the way, I think you forgot to add yourself as contributor of this module.

PR #536, opened 2025-12-16
[18.0][MIG] maintenance_purchase: Migration to version 18.0

christian-ramos-tecnativa:
Replaced the dependency on `maintenance_account` with `maintenance_product`, as the module works correctly without the accounting dependency
EDIT:
Also removed `purchase_stock` dependency to `purchase`
@Tecnativa TT59851
@pedrobaeza @victoralmau


pedrobaeza (commented):
Several comments:

- I don't think we need the `maintenance_product` dependency as well. It's not adding value.
- We should remove the flag at product level, as if you put a equipment category is enough for determining that you want to create an equipment. Why would you want to set an equipment category, but not to create an equipment? I know there's an automatism for setting the equipment category according the product category, but this is inherited from the other module, and we come back to the first point.
- Prepare the creation method to be inheritable for `maintenance_product_stock` for inhibiting the equipment creation when confirming, so better to refactor this creation into a separate method.

christian-ramos-tecnativa:
I think I get what you mean, but that would be a mayor refactor on how the module works right now, as instead of everything working by default configurations at product level using `maintenance_product` you will need to mannually set up the category everytime you create the purchase. Which IMO is a bad approach, unless we add a maintenance category at product level but in that case we are repeating configuration if the module `maintenance_product` is installed. 
I totally agree on making the equipment creation easier to extend and just saw the method `_prepare_account_move_line` that need to be also removed.

maintenance_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        items.write({"purchase_line_id": False, "purchase_id": False})
+        return super().unlink()
+
+    def action_view_equipments(self):
+        items = self.env["maintenance.equipment"].search(
+            [("purchase_id", "=", self.id)]
+        )
+        action = self.env.ref("maintenance.hr_equipment_action")
+        action_dict = action.sudo().read()[0]
```
victoralmau:
```suggestion
        action_dict = self.env["ir.actions.act_window"]._for_xml_id(
            "maintenance.hr_equipment_action"
        )        
```

maintenance_purchase/tests/test_purchase_order.py
```diff
@@ -0,0 +1,198 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import Command
+from odoo.tests.common import TransactionCase
+
+
+class TestMaintenanceProject(TransactionCase):
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class TestMaintenanceProject(BaseCommon):
```

maintenance_purchase/tests/test_purchase_order.py
```diff
@@ -0,0 +1,198 @@
...
+            }
+        )
+        cls.purchase_order = cls.PurchaseOrder.with_context(
+            tracking_disable=True
+        ).create(
+            {
+                "partner_id": cls.partner.id,
+            }
+        )
+        PurchaseOrderLine = cls.PurchaseOrderLine.with_context(tracking_disable=True)
```
victoralmau:
```suggestion
        PurchaseOrderLine = cls.PurchaseOrderLine
```

maintenance_purchase/views/purchase_order.xml
```diff
@@ -0,0 +1,33 @@
...
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[hasclass('oe_button_box')]" position="inside">
+                <button
+                    name="action_view_equipments"
+                    type="object"
+                    class="oe_stat_button"
+                    icon="fa-cubes"
+                    invisible="not equipment_count"
```
victoralmau, line 16:
```suggestion
                    invisible="not equipment_count"
                    groups="base_maintenance_group.group_maintenance_user"
```

maintenance_purchase/views/purchase_order.xml
```diff
@@ -0,0 +1,33 @@
...
+                        name="equipment_count"
+                        widget="statinfo"
+                    />
+                </button>
+            </xpath>
+            <xpath
+                expr="//notebook//field[@name='order_line']/list/field[@name='product_qty']"
+                position="before"
+            >
+                <field name="equipment_category_id" optional="hide" />
```
victoralmau, line 29:
```suggestion
                <field name="equipment_category_id" optional="hide" groups="base_maintenance_group.group_maintenance_user" />                
```

maintenance_purchase/views/maintenance_equipment.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record id="hr_equipment_view_form" model="ir.ui.view">
+        <field name="name">equipment.form</field>
+        <field name="model">maintenance.equipment</field>
+        <field name="inherit_id" ref="maintenance.hr_equipment_view_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//page[@name='product_information']//field[@name='serial_no']"
+                position="after"
+            >
+                <field name="purchase_id" />
```
victoralmau:
```suggestion
                <field name="purchase_id" groups="purchase.group_purchase_user" />
```

victoralmau (changes requested):
Code review OK

pedrobaeza:
> I think I get what you mean, but that would be a mayor refactor on how the module works right now, as instead of everything working by default configurations at product level using `maintenance_product` you will need to mannually set up the category everytime you create the purchase. Which IMO is a bad approach, unless we add a maintenance category at product level but in that case we are repeating configuration if the module `maintenance_product` is installed. I totally agree on making the equipment creation easier to extend and just saw the method `_prepare_account_move_line` that need to be also removed.

Right now is as well not direct, as even linking the equipment category in the product category, you need to put the check at product level, so it's very counter-intuitive (and the reason we weren't able to generate the equipment in our first tests), so I still think it's good to refactor it that way.

Changing the boolean at product level to be by default True is as well counter-intuitive, as it won't work unless the equipment category is put at product category, so both options are very bad combined.

You can keep the equipment category at product category level as default value for the purchase, but without depending on `maintenance_product`, and the way to not create the equipment is to empty the column, the same as in `account_asset_management`.

maintenance_purchase/views/purchase_order.xml
christian-ramos-tecnativa: I'm not sure about adding dependency to `base_maintenance_group` ngl

christian-ramos-tecnativa:
Instead of removing the dependency to `maintenance_product` as all the logic between product_category and equipment_category is there and changing the module behavior, what about adding: if we fill the equipments category at purchase line level force the creation of the equipment even if the product doesn't have the check active. This way we cover both options. WDYT @pedrobaeza

pedrobaeza:
I'm still seeing like "black magic" behind the curtains, which is confusing for the user, so I prefer to have it explicitly set.

guy-long-hair

christian-ramos-tecnativa:
No more "black magic" here. Now, if the category is removed, the equipment will not be created. If the category is set, the equipment will be created even if the product does not have the maintenance check.

maintenance_purchase/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Maintenance Purchase",
+    "summary": """
+        Create Equipments with purchases""",
+    "version": "18.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "CreuBlanca,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/maintenance",
+    "depends": [
+        "maintenance_product",
```
pedrobaeza: Why maintaining then this dependency? I want to be a stand-alone module.

christian-ramos-tecnativa: To retain part of the old behavior, the equipment category is auto-filled from the product category, to avoid setting it on every purchase line.

christian-ramos-tecnativa:
Added the posibility of changing the quantity to a higher value and creating the remaining equipments when re-confirming the purchase

maintenance_purchase/__manifest__.py
pedrobaeza: I prefer to not have it for not dragging that module that is half baked imo

maintenance_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,125 @@
...
+    _inherit = "purchase.order.line"
+
+    equipment_category_id = fields.Many2one(
+        comodel_name="maintenance.equipment.category",
+        string="Equipment Category",
+        compute="_compute_equipment_category_id",
+        store=True,
+        readonly=False,
+    )
+    equipment_ids = fields.Many2many(
```
pedrobaeza: Why is this a m2m instead of a o2m using `purchase_line_id` as inverse field?

maintenance_purchase/__manifest__.py
pedrobaeza: We should remove the dependency as stated.

pedrobaeza (changes requested):

christian-ramos-tecnativa:
Done! @pedrobaeza

maintenance_purchase/migrations/18.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+            UPDATE product_template pt
+            SET equipment_category_id = sub.equipment_category_id
+            FROM (
+                SELECT mec.product_category_id AS product_category_id,
+                        mec.id AS equipment_category_id
+                FROM maintenance_equipment_category mec
+                WHERE mec.product_category_id IS NOT NULL
+            ) AS sub
+            WHERE pt.categ_id = sub.product_category_id
+                AND pt.equipment_category_id IS NULL""",
```
pedrobaeza: You can short this SQL:

```suggestion
            SET equipment_category_id = mec.equipment_category_id
            FROM maintenance_equipment_category mec
            WHERE pt.categ_id = mec.product_category_id
                AND pt.equipment_category_id IS NULL""",
        )
```

maintenance_purchase/models/purchase_order.py
```diff
@@ -80,13 +78,11 @@ class PurchaseOrderLine(models.Model):
...
-                item.product_id.maintenance_ok
-                and item.product_id.product_tmpl_id.categ_id.equipment_category_ids
-            ):
-                item.equipment_category_id = fields.first(
-                    item.product_id.product_tmpl_id.categ_id.equipment_category_ids
+            if item.product_id.product_tmpl_id.equipment_category_id:
+                item.equipment_category_id = (
+                    item.product_id.product_tmpl_id.equipment_category_id
                 )
+
```
pedrobaeza: Remove this extra empty line

pedrobaeza (approved):
Almost there!

maintenance_purchase/models/product_template.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    equipment_category_id = fields.Many2one(
```
pedrobaeza: Maybe we should be more specific? `purchase_equipment_category_id`

christian-ramos-tecnativa:
Done!

pedrobaeza:
Well, I mean on the technical name, not the label, for not collapsing with other modules adding a field with the same name...

christian-ramos-tecnativa:
Done!

victoralmau (approved):
Code review OK

maintenance_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,122 @@
...
+    def _prepare_equipment_vals(self):
+        return {
+            "purchase_line_id": self.id,
+            "name": self.product_id.name,
+            "category_id": self.equipment_category_id.id,
+            "assign_date": self.order_id.date_order,
+            "effective_date": self.order_id.date_planned,
+            "partner_id": self.order_id.partner_id.id,
+            "partner_ref": self.order_id.partner_ref,
```
victoralmau, line 121:
```suggestion
            "partner_ref": self.order_id.partner_ref,
            "company_id": self.order_id.company_id.id,
```
I think `company_id` could be added (thus avoiding the need for `maintenance_purchase_stock` to add it).

christian-ramos-tecnativa: I add it in the other module because there is no way this gets executed without the company in the enviroment. But I also think is harmless so... Done!

PR #538, opened 2025-12-29
[18.0][FIX] maintenance_stock: invisible -> column invisible in list views

diggy128:
- Invisible only hid the values in the column
    - Replace invisible attribute with column_invisible in list views
    to hide the whole column when it should not be visible

### DIFF
--- a/maintenance_stock/views/stock_move_line_views.xml
+++ b/maintenance_stock/views/stock_move_line_views.xml
@@ -34,7 +34,7 @@
             <field name="product_id" position="after">
                 <field
                     name="maintenance_request_id"
-                    invisible="not context.get('show_maintenance_request_id', False)"
+                    column_invisible="not context.get('show_maintenance_request_id', False)"
                 />
             </field>
         </field>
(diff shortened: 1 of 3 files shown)


maintenance_security/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "website": "https://github.com/OCA/maintenance",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "depends": ["maintenance"],
+    "depends": ["hr", "maintenance"],
```
victoralmau:
```suggestion
    "depends": ["maintenance"],
```
IMO `hr` dependency is not necessary.

victoralmau (changes requested):

maintenance_security/__manifest__.py
diggy128: True, the requiring entry was split to module hr_maintenance_security.

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #539, opened 2025-12-31
[18.0][ADD] maintenance_purchase_stock: New module

christian-ramos-tecnativa:
Depends on:

- [x] https://github.com/OCA/maintenance/pull/536

@Tecnativa TT59744
@pedrobaeza @victoralmau


dalonsod:
Hello @christian-ramos-tecnativa only one suggestion: could you consider renaming this addon? At this repo there's already an addon called `maintenance_stock`, and this new could suggest a link between this one and `maintenance_purchase`, and this could lead to some confusion, what do you think about it?

christian-ramos-tecnativa:
Hi @dalonsod, I see your point. However, in cases like this, users should refer to the module description to understand its purpose. For example, the maintenance_stock module handles consumptions, not a full integration with stock.

christian-ramos-tecnativa:
Ready for review/merge

maintenance_purchase_stock/models/purchase_order.py
```diff
@@ -0,0 +1,28 @@
...
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    def _qty_to_create_equipments(self):
+        return int(self.qty_received - len(self.equipment_ids))
+
+    @api.depends("move_ids.state", "move_ids.product_uom", "move_ids.quantity")
+    def _compute_qty_received(self):
+        res = super()._compute_qty_received()
+        self.order_id._create_equipments()
```
pedrobaeza, line 27: This shouldn't be done on a computed field, as this computation may be done as a result of a `onchange`, and later discarded, so please intercept instead `write`.

pedrobaeza: ping @christian-ramos-tecnativa

christian-ramos-tecnativa: This compute has the decorator `@api.depends` so it should be triggered when writing any of the fields above, not on UI change. Is there any situation that can trigger it as an onchange?

pedrobaeza: @christian-ramos-tecnativa but this is triggered not when writing, but by web client in any UI change of one of the depends, but it can be later dismissed if you press "Discard" button, having the created equipments already there. As said, this should be done on write.

PR #541, opened 2026-01-19
[19.0][MIG] maintenance_product: Migration to 19.0 

peluko00:
Migration to v19

cc https://github.com/APSL 32948
@miquelalzanillas @javierobcn @mpascuall @BernatObrador @ppyczko @lbarry-apsl please review


mpascuall (approved):
LGTM!

ppyczko (approved):
LGTM!

maintenance_product/models/maintenance_equipment.py
```diff
@@ -30,7 +30,7 @@ def _onchange_product_id(self):
             self.name = self.product_id.name
             self.cost = self.product_id.standard_price
             if self.product_id.seller_ids:
-                first_seller = fields.first(self.product_id.seller_ids)
+                first_seller = self.product_id.seller_ids[0]
```
BhaveshHeliconia, line 33: In Odoo 19, fields.first() is no longer used, and the standard approach is to use next(iter(...), ...) instead.

Instead of accessing the first record with [0], it would be better to follow the Odoo 19 pattern for consistency, e.g.: first_seller = next(iter(self.product_id.seller_ids), False)

This aligns with the current Odoo standard and avoids direct indexing on recordsets.

BhaveshHeliconia (changes requested):

maintenance_product/models/maintenance_equipment.py
pedrobaeza: Using `[:1]` is also a good shorter alternative.

pedrobaeza: Anyway in this case that there's an if checking if there's a value, also `[0]` is good enough, as now there's no advantages using `fields.first` / `next(iter` than this, as the cache is since a lot of versions not related to the recordset, but global.

BhaveshHeliconia: Agree that in this case, since we already check for existence, using [0] is fine.

BhaveshHeliconia (approved):

PR #547, opened 2026-03-19
[18.0][ADD] maintenance_helpdesk_mgmt: New module

christian-ramos-tecnativa:
New module to link maintenance requests with helpdesk tickets

@Tecnativa TT61468
@pedrobaeza @cristina-hidalgo-tecnativa

PR #549, opened 2026-03-23
[18.0][IMP] maintenance_helpdesk_mgmt: Added kanban actions and searchviews

christian-ramos-tecnativa:
Added comments from @dalonsod to improve the new module added in:

- https://github.com/OCA/maintenance/pull/547
- Filters `With helpdeskt tickets` and `With maintenance requests`
- Actions to display and open records from kanban views 

Thanks for the review!
@Tecnativa TT61468
@pedrobaeza @david-banon-tecnativa


maintenance_helpdesk_mgmt/models/maintenance_request.py
```diff
@@ -14,3 +14,20 @@ class MaintenanceRequest(models.Model):
         string="Helpdesk Tickets",
         copy=False,
     )
+    helpdesk_tickets_count = fields.Integer(compute="_compute_helpdesk_tickets_count")
+
+    def _compute_helpdesk_tickets_count(self):
+        for record in self:
+            record.helpdesk_tickets_count = len(record.helpdesk_ticket_ids)
```
dalonsod: In order to improve performance, you could use `read_group` instead of `len`, because count field is used across `kanban` view.

The same could be applied to `maintenance_requests_count`

maintenance_helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -24,4 +24,39 @@
...
+        <field name="inherit_id" ref="helpdesk_mgmt.view_helpdesk_ticket_kanban" />
+        <field name="arch" type="xml">
+            <field name="assigned_date" position="after">
+                <field name="maintenance_request_ids" />
+                <field name="maintenance_requests_count" />
+            </field>
+            <field name="tag_ids" position="after">
+                <div class="row" t-if="record.maintenance_requests_count.value > 0">
+                    <a type="object" name="action_view_maintenance_request">
+                        Maintenance Requests <field name="maintenance_requests_count" />
```
dalonsod: Is it possible to switch legend and count number? Like e.g. "2 Maintenance Requests"

Image

maintenance_helpdesk_mgmt/views/helpdesk_ticket_views.xml
```diff
@@ -24,4 +24,39 @@
...
         </field>
     </record>
+
+    <record id="view_helpdesk_ticket_kanban" model="ir.ui.view">
+        <field name="name">helpdesk.ticket.kanban</field>
+        <field name="model">helpdesk.ticket</field>
+        <field name="inherit_id" ref="helpdesk_mgmt.view_helpdesk_ticket_kanban" />
+        <field name="arch" type="xml">
+            <field name="assigned_date" position="after">
+                <field name="maintenance_request_ids" />
```
dalonsod: Is this field really needed at this view?

dalonsod (changes requested):
Code and functional review

christian-ramos-tecnativa:
Every comment were addressed, Thanks again for the review!

maintenance_helpdesk_mgmt/models/maintenance_request.py
```diff
@@ -14,3 +14,30 @@ class MaintenanceRequest(models.Model):
         string="Helpdesk Tickets",
         copy=False,
     )
+    helpdesk_tickets_count = fields.Integer(compute="_compute_helpdesk_tickets_count")
```
pedrobaeza: Please avoid plurals in the middle of the field name

christian-ramos-tecnativa: Done, thanks!

pedrobaeza (approved):

dalonsod (approved):
Tested at runboat 👍 

There's a minor change (adding security access control to ticket page at request form) that could be placed at a standalone `FIX` commit, but I think it's not very important for a newly created module, like this one.

PR #550, opened 2026-04-08
[18.0][FIX] maintenance_plan: default company in requests

LauraCForgeFlow:
The field `company_id` is not required on `maintenance.equipment` or `maintenance.plan` (to support multi-company shared records), but it is required on `maintenance.request`. When generating requests via `_prepare_request_from_plan`, the `company_id` field evaluates to False if neither the plan nor the equipment has a company set. This causes an error because company is a required field on maintenance.request.

To fix this, we add a final fallback, which is consistent with how `maintenance.request` itself defaults its `company_id`. This ensures a valid company is always provided without changing behavior when the plan or equipment already has one. Also added some testing to cover the use case.

### DIFF
--- a/maintenance_plan/models/maintenance_equipment.py
+++ b/maintenance_plan/models/maintenance_equipment.py
@@ -92,6 +92,7 @@ class MaintenanceEquipment(models.Model):
             description=description,
         )
 
+        req_company = maintenance_plan.company_id or self.company_id
         data = {
             "name": name,
             "request_date": next_maintenance_date,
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (approved):
👍  Code LGTM

MarinaAForgeFlow (approved):

LoisRForgeFlow (approved):
👍

maintenance_plan/models/maintenance_equipment.py
```diff
@@ -92,6 +92,7 @@ def _prepare_request_from_plan(self, maintenance_plan, next_maintenance_date):
             description=description,
         )
 
+        req_company = maintenance_plan.company_id.id or self.company_id.id
```
pedrobaeza: This is an ID, so name properly the variable:


```suggestion
        req_company_id = maintenance_plan.company_id.id or self.company_id.id
```

Anyway, it's better to store the recordset if already browsed, for any other usage that can be done in the future in the method:


```suggestion
        req_company = maintenance_plan.company_id or self.company_id
```

pedrobaeza (approved):
Thanks!

/ocabot merge patch

PR #552, opened 2026-04-09
[17.0][BACKPORT] maintenance_purchase: Backport from 18.0 to 17.0

vvrossem:


yankinmax (approved):
Thank you

jcoux (approved):
Technically seems good 👍

maintenance_purchase/migrations/18.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2026 Tecnativa - Christian Ramos
```
BhaveshHeliconia, line 1: We need to remove this migration directory since it is not related to this version.

maintenance_purchase/tests/test_purchase_order.py
```diff
@@ -0,0 +1,164 @@
...
+        cls.ProductCategory = cls.env["product.category"]
+        cls.ProductProduct = cls.env["product.product"]
+        cls.ProductTemplate = cls.env["product.template"]
+        cls.ResPartner = cls.env["res.partner"]
+
+        cls.partner = cls.ResPartner.create(
+            {
+                "name": "partner",
+            }
+        )
```
BhaveshHeliconia, line 27:
```suggestion
```
No need to create a res.partner record when using BaseCommon, as it is already provided by the base class.

BhaveshHeliconia:
also could you please squash the administrative and bot/weblate commits?

PR #564, opened 2026-05-20
[FIX] maintenance_equipment_sequence: convert create to api.model_create_multi

miquelrosell99:
Converts MaintenanceEquipmentCategory.create() and MaintenanceEquipment.create() to batch-compatible api.model_create_multi for Odoo 17 best practice.

### DIFF
--- a/maintenance_equipment_sequence/models/maintenance.py
+++ b/maintenance_equipment_sequence/models/maintenance.py
@@ -63,16 +63,23 @@ class MaintenanceEquipmentCategory(models.Model):
                 sequence = category.sequence_id._get_current_sequence()
                 sequence.sudo().number_next = category.sequence_number_next
 
-    @api.model
-    def create(self, vals):
-        if not vals.get("sequence_id", False):
-            if vals.get("sequence_prefix", False):
-                vals["sequence_id"] = self.sudo()._create_sequence(vals).id
-        else:
-            vals["sequence_prefix"] = (
-                self.env["ir.sequence"].browse(vals["sequence_id"]).prefix
-            )
-        result = super().create(vals)
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("sequence_id"):
+                if vals.get("sequence_prefix"):
+                    vals["sequence_id"] = self.sudo()._create_sequence(vals).id
+        # Batch-read prefixes for provided sequence IDs
+        sequence_ids = {
+            vals["sequence_id"] for vals in vals_list if vals.get("sequence_id")
+        }
+        if sequence_ids:
+            sequences = self.env["ir.sequence"].browse(list(sequence_ids))
+            prefix_map = {seq.id: seq.prefix for seq in sequences}
+            for vals in vals_list:
+                if vals.get("sequence_id"):
+                    vals["sequence_prefix"] = prefix_map.get(vals["sequence_id"])
+        result = super().create(vals_list)
         self._compute_equipment_code()
         return result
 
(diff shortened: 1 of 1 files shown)


maintenance_equipment_sequence/models/maintenance.py
```diff
@@ -63,16 +63,17 @@ def _inverse_seq_number_next(self):
...
-        result = super().create(vals)
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("sequence_id", False):
+                if vals.get("sequence_prefix", False):
+                    vals["sequence_id"] = self.sudo()._create_sequence(vals).id
+            else:
+                vals["sequence_prefix"] = (
+                    self.env["ir.sequence"].browse(vals["sequence_id"]).prefix
```
dalonsod: This line implies a `browse()` access for each `for` iteration. IMO it is desirable moving it outside `for` statement, make only one `browse` and manage result with a proper python dict within `for`.

maintenance_equipment_sequence/models/maintenance.py
```diff
@@ -107,18 +108,19 @@ def _compute_equipment_code(self):
...
-                equipment.serial_no = sequence_id._next()
-        return equipment
+    @api.model_create_multi
+    def create(self, vals_list):
+        equipments = super().create(vals_list)
+        for equipment, vals in zip(equipments, vals_list):
+            if equipment.category_id and not equipment.serial_no:
+                sequence_id = (
+                    self.env["maintenance.equipment.category"]
+                    .browse(vals.get("category_id"))
```
dalonsod: The same as above

dalonsod (changes requested):
See possible performance improvement comment

maintenance_equipment_sequence/models/maintenance.py
miquelrosell99: Done, thanks!

maintenance_equipment_sequence/models/maintenance.py
miquelrosell99: Done, thanks!

miquelrosell99:
Hi @dalonsod, thanks for the review.

The batching change you suggested is already in place in the current branch:

- In `MaintenanceEquipmentCategory.create`, the `ir.sequence` records are browsed once and mapped by ID before looping over `vals_list`.
- In `MaintenanceEquipment.create`, the `maintenance.equipment.category` records are browsed once and mapped to their sequence before looping over the created equipments.

I believe this addresses the performance concern. Could you please re-review when you have a moment?

maintenance_equipment_sequence/models/maintenance.py
```diff
@@ -63,16 +63,23 @@ def _inverse_seq_number_next(self):
...
+            vals["sequence_id"] for vals in vals_list if vals.get("sequence_id")
+        }
+        if sequence_ids:
+            sequences = self.env["ir.sequence"].browse(list(sequence_ids))
+            prefix_map = {seq.id: seq.prefix for seq in sequences}
+            for vals in vals_list:
+                if vals.get("sequence_id"):
+                    vals["sequence_prefix"] = prefix_map.get(vals["sequence_id"])
+        result = super().create(vals_list)
         self._compute_equipment_code()
```
cgarcia-solvos: `_compute_equipment_code` should be called on `result`, which holds the newly created records `self` at this point is an empty recordset. Likewise, the outcome is the same, since a newly created category can never have associated equipment yet,  the method would always be a no-op.

```suggestion
        result._compute_equipment_code()
```

cgarcia-solvos (approved):
Code review. LGTM

PR #570, opened 2026-07-04
[18.0][IMP] maintenance_helpdesk_mgmt: Explicitly set internal user on ticket fields

pedrobaeza:
We assume that anyone accessing to the backend always belongs to the "internal user" group, but that may not be true, for example, in the standard project external sharing, and now, with the module `base_group_subcontractor`, if giving permissions to helpdesk tickets.

Thus, let's explicitly assign the internal user group to the ticket extensions for avoiding errors if we only want to access this part.

@Tecnativa TT63310


eduezerouali-tecnativa (approved):
LGTM

BhaveshHeliconia (approved):

PR #574, opened 2026-07-10
[18.0][ADD] maintenance_equipment_tier_validation

smorita7749:
[QT6788](https://www.quartile.co/web#id=6788&cids=3&model=project.task&view_type=form)


maintenance_equipment_tier_validation/models/maintenance_equipment.py
```diff
@@ -0,0 +1,35 @@
...
+    _state_to = ["active"]
+    _cancel_state = "inactive"
+
+    _tier_validation_manual_config = False
+
+    state = fields.Selection(
+        selection=[
+            ("draft", "Draft"),
+            ("active", "Active"),
+            ("inactive", "Inactive"),
```
AungKoKoLin1997: Seems like these three attributes are not matching. Shouldn't it be like draft, confirmed, cancel?
Active and Inactive are part of active field, no?

maintenance_equipment_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "summary": "Support a tier validation process for maintenance equipment",
+    "version": "18.0.1.0.0",
+    "category": "Maintenance",
+    "website": "https://github.com/OCA/maintenance",
+    "author": "Quartile, Odoo Community Association (OCA)",
+    "maintainers": ["smorita7749"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["base_maintenance", "base_tier_validation"],
```
AungKoKoLin1997: I am not comfortable adding `base_maintenance` as a dependency just for `header` attribute in the equipment form view. I don't have the clean solution yet but I have in mind now is using `_get_view()`.
Please let me know if you have better approach.

```
@api.model
    def _get_view(self, view_id=None, view_type=False, **options):
        arch, view = super()._get_view(view_id, view_type, **options)
        if view_type == "form":
            header = arch.find(".//header")
            if header is None: 
                sheet = arch.find(".//sheet")
                if sheet is not None:
                    header = etree.Element("header")
                    sheet.addprevious(header)
            if header is not None:
                button = etree.SubElement(header, "button")
                button.set("name", "action_do_something")
                button.set("type", "object")
                button.set("string", "Do Something")
        return arch, view
```

maintenance_equipment_tier_validation/models/maintenance_equipment.py
smorita7749: This is what I am concerned about.
Ideally, I think it would be better to handle this in a separate module (as is the case with other Tier validation systems), but for now, I am considering managing the status within this module. (using "Draft" before approval, and "Active" or "Inactive" after approval).

Should I implement a change to include a unique code for the status?
https://github.com/OCA/maintenance/tree/18.0/maintenance_equipment_status

AungKoKoLin1997: I don't think adding a unique code for the status is not the solution because `base_tier_validation` doesn't work on many2one field. So, adding state field in this module is valid for me but the attributes should be `draft`, `confirmed`, `cancel`.

smorita7749: Done!

oca_dependencies.txt
AungKoKoLin1997, line 1: Is it auto generated by running pre-commit?

maintenance_equipment_tier_validation/__manifest__.py
smorita7749: I chose you suggestion.

oca_dependencies.txt
smorita7749: I removed it

maintenance_equipment_tier_validation/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
AungKoKoLin1997, line 2:
```suggestion
# Copyright 2026 Quartile
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

maintenance_equipment_tier_validation/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
AungKoKoLin1997, line 2:
```suggestion
# Copyright 2026 Quartile
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

README.md
```diff
@@ -32,6 +32,7 @@ addon | version | maintainers | summary
 [maintenance_equipment_sequence](maintenance_equipment_sequence/) | 18.0.1.0.0 |  | Adds sequence to maintenance equipment defined in the equipment's category
 [maintenance_equipment_status](maintenance_equipment_status/) | 18.0.1.0.0 |  | Maintenance Equipment Status
 [maintenance_equipment_tags](maintenance_equipment_tags/) | 18.0.1.0.0 | <a href='https://github.com/etobella'><img src='https://github.com/etobella.png' width='32' height='32' style='border-radius:50%;' alt='etobella'/></a> | Adds category tags to equipment
+[maintenance_equipment_tier_validation](maintenance_equipment_tier_validation/) | 18.0.1.0.0 |  | Support a tier validation process for maintenance equipment
```
AungKoKoLin1997: Is this change part of the PR?

AungKoKoLin1997 (approved):
Code Review: LG

Minor points.

maintenance_equipment_tier_validation/models/__init__.py
smorita7749: Done!

maintenance_equipment_tier_validation/__init__.py
smorita7749: Done!

README.md
smorita7749: I removed it.

BhaveshHeliconia (approved):
Code and functional review LGTM!

PR #578, opened 2026-07-23
[18.0][ADD] maintenance_request_done_readonly

smorita7749:
This module adds the restriction for editing by general users after maintenance.request is completed and allows only maintenance manager to edit it.

@qrtl QT6788


maintenance_request_done_readonly/models/res_config_settings.py
```diff
@@ -0,0 +1,38 @@
...
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    maintenance_done_editable_field_ids = fields.Many2many(
+        "ir.model.fields",
+        string="Editable fields on completed requests",
+        domain=[("model", "=", "maintenance.request"), ("store", "=", True)],
+        help="Fields that non-managers may still edit after a maintenance "
+        "request is completed. Other fields stay locked.",
+    )
```
AungKoKoLin1997: It would be better if we use system parameter for that because it is a model specific and we don't have a plan to have a different settings depends on company.

maintenance_request_done_readonly/models/maintenance_request.py
```diff
@@ -0,0 +1,65 @@
...
+            "message_follower_ids",
+            "message_ids",
+            "activity_ids",
+        }
+        names = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("maintenance_request_done_readonly.editable_fields", "")
+        )
+        return allowed | {name for name in names.split(",") if name}
```
AungKoKoLin1997: I think this method is mixed with both hardcoded fields list and system parameter. Shouldn't we do like below and the hardcoded fields should be added in system parameter if it is needed. How do you think?

```
 def _get_write_bypass_fields(self):
        param = (
            self.env["ir.config_parameter"]
            .sudo()
            .get_param("maintenance_request_done_readonly.write_bypass_fields", "")
        )
        return {f.strip() for f in param.split(",") if f.strip()}
```

smorita7749: I originally hardcoded these fields to keep chatter always writable. But I verified chatter/activities don't go through this guarded write(), so the hardcoded list isn't needed and moved everything to the parameter as suggested.

maintenance_request_done_readonly/models/maintenance_request.py
```diff
@@ -0,0 +1,49 @@
...
+        ):
+            locked = self.filtered("done")
+            if locked:
+                raise UserError(
+                    _(
+                        "'%s' is completed and can only be edited by a "
+                        "maintenance manager.",
+                        locked[0].display_name,
+                    )
+                )
```
AungKoKoLin1997, line 47: How about like that?
```suggestion
        if (
            not self.env.context.get("mnt_done_bypass_lock")
            and not self.env.user.has_group("maintenance.group_equipment_manager")
            and (set(vals) - self._writable_fields_when_done())
            and (locked := self.filtered("done"))
        ):
            raise UserError(
                _(
                    "'%s' is completed and can only be edited by a "
                    "maintenance manager.",
                    locked[0].display_name,
                )
            )
```

maintenance_request_done_readonly/models/maintenance_request.py
smorita7749: I found that hardcoding `message_main_attachment_id` can be acceptable because users cannot comment with an attachment if they exclude `message_main_attachment_id` from  maintenance_request_done_readonly.write_bypass_fields.
What do you think whether hardcoding or having it in data.xml?

maintenance_request_done_readonly/models/maintenance_request.py
smorita7749: Updated!
Also, I added a comment about the subsequent process.

maintenance_request_done_readonly/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,15 @@
+This module makes completed maintenance requests read-only.
+
+Once a request reaches a done stage, it can no longer be edited or reopened:
+any attempt to change its fields (including moving it back out of the done
+stage) is blocked. Only members of the *Equipment Manager* group keep full
+access to completed requests.
```
AungKoKoLin1997: Just a thought. Shouldn't we create new group instead of relying on manager?
With current design, normal user can't be able to edit.

AungKoKoLin1997 (approved):
Code Review: LGTM

maintenance_request_done_readonly/readme/DESCRIPTION.md
smorita7749: I added a new group to edit done request.

maintenance_request_done_readonly/security/res_groups.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="group_edit_done_request" model="res.groups">
```
AungKoKoLin1997:
```suggestion
    <record id="group_maintenance_request_edit_done" model="res.groups">
```

maintenance_request_done_readonly/security/res_groups.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="name">Edit Completed Maintenance Requests</field>
```
AungKoKoLin1997:
```suggestion
        <field name="name">Maintenance: Edit Completed Requests</field>
```

maintenance_request_done_readonly/security/res_groups.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field
+            name="category_id"
+            ref="base.module_category_manufacturing_maintenance"
```
AungKoKoLin1997: I think using that category can add new maintenance label in the res.user view or appears in the drop down list. It should be `base.module_category_extra_rights` in my opinion.

maintenance_request_done_readonly/security/res_groups.xml
smorita7749: Done!

maintenance_request_done_readonly/security/res_groups.xml
smorita7749: Done!

maintenance_request_done_readonly/security/res_groups.xml
smorita7749: Could you please take a look if my understanding is correct?

AungKoKoLin1997 (approved):
LGTM

BhaveshHeliconia (approved):
Code and functional review LGTM!

maintenance_request_done_readonly/security/res_groups.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="implied_ids" eval="[(4, ref('base.group_user'))]" />
+        <field
+            name="comment"
+        >The user will be able to edit completed maintenance requests.</field>
+    </record>
+    <!-- Equipment managers keep editing completed requests, as before. -->
+    <record id="maintenance.group_equipment_manager" model="res.groups">
+        <field
+            name="implied_ids"
+            eval="[(4, ref('group_maintenance_request_edit_done'))]"
```
yostashiro, line 15: Use Command.link

maintenance_request_done_readonly/models/maintenance_request.py
yostashiro, line 1: Can you please check if we should add @api.ondelete method to block unlink operations on done requests.

maintenance_request_done_readonly/__manifest__.py
yostashiro, line 1: Please add yourself as a maintainer.

smorita7749:
What I changed:
- Use `Command.link` in `security/res_groups.xml`.
- Add `_unlink_except_done` to block deleting completed requests for users outside the edit group, with tests.
- Add myself as a maintainer.

maintenance_request_done_readonly/security/res_groups.xml
smorita7749: Done!

maintenance_request_done_readonly/models/maintenance_request.py
smorita7749: I think it is necessary.
I changed to only the Maintenance: Edit Completed Requests group can delete done requests.

maintenance_request_done_readonly/__manifest__.py
smorita7749: Done!

PR #580, opened 2026-07-24
[19.0][FIX] maintenance_plan: post_init_hook crashes on Odoo 19 (removed equipment fields)

carlosffp:
Installing `maintenance_plan` on Odoo 19 fails during `post_init_hook` with:

```
AttributeError: 'maintenance.equipment' object has no attribute 'next_action_date'
```

The migration hook references two `maintenance.equipment` fields that were **removed in Odoo 19 core**:

- `next_action_date` — used to match "the" open preventive request by date
- `mtbf` — used as the plan `duration`

Because it runs in `post_init_hook`, the whole install transaction rolls back and the module cannot be installed on any database that has equipments with `expected_mtbf` set.

### Fix
Port the hook to Odoo 19:

- Derive the plan `interval` from `expected_mtbf` (in days) and set `interval_step = "day"`.
- Drop the `next_action_date` matching / `UserError` branch (there is no next action date field to match against in 19) and the `duration = mtbf` line.
- Batch the `maintenance.plan` creation.

The end result is equivalent: one `maintenance.plan` (kind "Install") per equipment that had an expected MTBF, using the MTBF as the interval.

Tested by installing the module on an Odoo 19 database (ASSA) with equipments that have `expected_mtbf` set — install now completes and creates the expected plans.

### DIFF
--- a/maintenance_plan/README.rst
+++ b/maintenance_plan/README.rst
@@ -47,10 +47,10 @@ Despite Odoo's built-in recurrence functionality included in the
 maintenance requests themselves (since 17.0), this module could be
 needed for some reasons:
 
-- Odoo implementation is javascript only
-- Some companies already work with the existing "Maintenance Plan"
-  workflow
-- Some modules are depending on this one
+-  Odoo implementation is javascript only
+-  Some companies already work with the existing "Maintenance Plan"
+   workflow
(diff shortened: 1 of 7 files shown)


BhaveshHeliconia (approved):
Thanks for the fix!

PR #548, opened 2026-03-20
[19.0][MIG] maintenance_equipment_certification

JasminSForgeFlow:
Standard Migration

Below things are improved in module, not a part of migration

- Add certificate number and renewal date
- Add dedicated menu for certificate
- Send an email notification when a certificate is about to expire
- Configure how many days before need to send email, it can be multiple ( i.e. 20 days before, 10 days before)
- Add active/inactive feature

@ForgeFlow

PR #591, opened 2026-09-09
[MIG] maintenance_equipment_certification: Migration to 18.0

SirPyTech:
Backport from `19.0` (https://github.com/OCA/maintenance/tree/de2cd52ab404f8127d3f060af62cfc8d8613b431/maintenance_equipment_certification) to include the awesome improvements added in https://github.com/OCA/maintenance/pull/548 from @JasminSForgeFlow.

In the migration I updated part of the documentation in order to correspond to what is shown to the user, it should probably be updated in `19.0` too.

I also added a small fix in the last commit because the archived view was showing like this: image
after this change, it is image.


HekkiMelody (approved):
Code review, LGTM

PR #364, opened 2023-10-11
[IMP] maintenance_equipment_contract: Access requests from contract in order to verify that all has been processed

etobella:
Usually, users want to see that the supplier managed all the preventive requests when managing an invoice from a contract.

PR #594, opened 2026-09-21
[18.0][IMP] maintenance_equipment_contract: Access requests from contract in order to verify that all has been processed

deeniiz:
Forward port of #364, #170 and #222 from 14.0


deeniiz:
@eantones , can you please review it?

maintenance_equipment_contract/security/contract_security.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="equipment_contract_see_all" model="ir.rule">
+        <field name="name">See All Contracts</field>
+        <field name="model_id" ref="contract.model_contract_contract" />
+        <field name="domain_force">[("equipment_ids",'!=',False)]</field>
+        <field
+            name="groups"
+            eval="[
+            (4,ref('maintenance.group_equipment_manager'))]"
```
dalonsod, line 10: Use `Command.link` instead

maintenance_equipment_contract/views/contract_contract.xml
```diff
@@ -9,12 +9,23 @@
...
+                    invisible="equipment_count == 0"
+                />
+            </div>
             <field name="tag_ids" position="before">
                 <field
                     name="equipment_ids"
                     widget="many2many_tags"
                     options="{'no_create_edit': True}"
                 />
+                <field name="equipment_count" invisible="1" />
```
dalonsod, line 28: For `18.0` this field is no longer required, as Odoo will automatically add it to view.

maintenance_equipment_contract/models/contract_contract.py
```diff
@@ -1,10 +1,27 @@
...
 class ContractContract(models.Model):
     _inherit = "contract.contract"
 
     equipment_ids = fields.Many2many("maintenance.equipment", string="Equipments")
+    equipment_count = fields.Integer(compute="_compute_equipment_count")
+
+    @api.depends("equipment_ids")
+    def _compute_equipment_count(self):
+        for record in self:
+            record.equipment_count = len(record.equipment_ids)
```
dalonsod, line 16: This is right and will work in a normal scenario. Anyway, a good and optional improvement here could be moved to a `search_count` or something like that. Once done, this field could be securely added to e.g. `list` view, without lack of performance.

dalonsod (changes requested):
Only code review, see comments

